PR #18, opened 2019-03-05
Added LicenseCard component

doytch:
# LicenseCard

Renders a card displaying information about licenses of the type [`License`](https://github.com/folio-org/mod-licenses/blob/master/service/grails-app/domain/org/olf/licenses/License.groovy) implemented by the `licenses` Okapi interface.

## Basic Usage
Simply passing a license object is all that's required.
```
import { LicenseCard } from '@folio/stripes-erm-components';
...
return <LicenseCard license={license} />;
```

## Props

| Name | Type | Description | Required | Default |
--- | --- | --- | --- | --- |
| `className` | string |
| `license` | object | [`License`](https://github.com/folio-org/mod-licenses/blob/master/service/grails-app/domain/org/olf/licenses/License.groovy) | Yes |
| `renderName` | bool | Set to `false` to turn off rendering the license name  | | `true` |

### DIFF
--- a/lib/DocumentsFieldArray/DocumentsFieldArray.js
+++ b/lib/DocumentsFieldArray/DocumentsFieldArray.js
@@ -69,7 +69,7 @@ export default class DocumentsFieldArray extends React.Component {
     const { fields } = this.props;
 
     return docs.map((doc, i) => (
-      <div className={css.doc}>
+      <div className={css.doc} key={i}>
         <Row>
           <Col xs={11}>
             <Field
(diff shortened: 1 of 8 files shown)


lib/DocumentsFieldArray/DocumentsFieldArray.js
```diff
@@ -69,7 +69,7 @@ export default class DocumentsFieldArray extends React.Component {
     const { fields } = this.props;
 
     return docs.map((doc, i) => (
-      <div className={css.doc}>
+      <div className={css.doc} key={i}>
```
doytch: Just rolled this separate fix into this PR

PR #26, opened 2019-03-07
Add support for qindex

MikeTaylor:
I need this so that when the user chooses to search in an index other than the default one specified as `searchKey` at compile-time, that selection can be overridden by the `qindex` portion of the URL query.

### DIFF
--- a/lib/getSASParams.js
+++ b/lib/getSASParams.js
@@ -10,10 +10,10 @@ export default (options) => (queryParams, pathComponents, resources) => {
     stats: 'true',
   };
 
-  const { query: { query, filters, sort } } = resources;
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Thanks, Mark. Will you please let me know when there's a release of `stripes-erm-components` that includes this?

doytch:
For sure @MikeTaylor . I'm assuming you're not using the folioci registry and want to know about a release on the official folio registry?

MikeTaylor:
Oh, that's a good point! Yes, I meant on the official registry, but I do indeed have the option of using folioci.

However, annoyingly, it looks like I'm going to need to make another minor change. So let's get that done and PR'd before we get too carried away.

PR #35, opened 2019-03-19
Erm 96: Create new LicenseTermsList

CalamityC:
* add directory for new LicenseTermsList component
* add LicenseTermsList.js, readme.md
* update index.js, package.json, CHANGELOG.md, translation files

### DIFF
--- a/lib/LicenseTermsList/LicenseTermsList.js
+++ b/lib/LicenseTermsList/LicenseTermsList.js
@@ -55,7 +55,7 @@ export default class LicenseTerms extends React.Component {
 
   render() {
     const { id, onToggle, open, parentResources } = this.props;
-    const terms = get(parentResources.terms, ['records'], []);
+    const terms = get(parentResources, ['terms', 'records'], []);
 
     return (
       <Accordion
--- a/lib/LicenseTermsList/LicenseTermsList.js
+++ b/lib/LicenseTermsList/LicenseTermsList.js
@@ -3,34 +3,28 @@ import PropTypes from 'prop-types';
 import { get } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import {
-  Accordion,
   Col,
   InfoPopover,
   Layout,
   Row
 } from '@folio/stripes/components';
 
-export default class LicenseTerms extends React.Component {
+export default class LicenseTermsList extends React.Component {
   static propTypes = {
     license: PropTypes.shape({ customProperties: PropTypes.object }),
-    id: PropTypes.string,
-    onToggle: PropTypes.func,
-    open: PropTypes.bool,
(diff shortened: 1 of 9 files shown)


translations/stripes-erm-components/de.json
```diff
@@ -20,5 +20,6 @@
     "licenseCard.status": "Status",
     "licenseCard.startDate": "Start date",
     "licenseCard.endDate": "End date",
-    "licenseCard.licensor": "Licensor"
+    "licenseCard.licensor": "Licensor",
+    "licenseTermsList.notSet": "Not set"
 }
```
dltj: These changes are not needed.  (In fact, changes to translation files other than `en.json` will be automatically overwritten when the Lokalise-to-GitHub script is run.)  In other words, these changes are harmless, but they will not "stick".

lib/LicenseTermsList/readme.md
```diff
@@ -0,0 +1,17 @@
+# LicenseTermsList
+
+
+
+
+## Basic Usage
+Simply passing a license object and it's terms is all that's required.
```
doytch: Can you provide provide more info about the `terms` array and its structure in the **Basic Usage** and **Props** sections? I try to write these readmes so it's possible to reference the readme rather than having to look at the code when using the components, and I'd have no idea what the terms array is supposed to look like from this.

(And also remove the extra blank lines after the header?)

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -0,0 +1,57 @@
...
+  InfoPopover,
+  Layout,
+  Row
+} from '@folio/stripes/components';
+
+export default class LicenseTermsList extends React.Component {
+  static propTypes = {
+    license: PropTypes.shape({ customProperties: PropTypes.object }),
+    terms: PropTypes.arrayOf(PropTypes.shape({
+      primary: PropTypes.bool,
```
doytch: Can you alphabetically sort props whenever we have a list like this?

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -0,0 +1,57 @@
...
+          <Col xs={12} md={8}>
+            <span data-test-term-value={term.name}>{value}</span>
+          </Col>
+        </Row>
+      </Layout>
+    );
+  }
+
+  render() {
+    const { terms } = this.props;
```
doytch: Let's write this a little simpler like `return this.props.terms.map(this.renderTerm)`

doytch:
Still the documentation stuff, but Github warning about a conflict made me realise that you need to rebase this branch to master and then bump the version of this module to `1.2.1` since we've already released `1.2.0`.

PR #39, opened 2019-04-04
ERM-79: set supplementary information for a license

CalamityC:
* modify DocumentCard and DocumentFieldArray components to support supplementary information with document categories
* add minor version 1.3.1 to CHANGELOG.md and package.json
* add doc.category to en.json
* add .DS_Store to .gitignore

### DIFF
--- a/lib/DocumentCard/DocumentCard.js
+++ b/lib/DocumentCard/DocumentCard.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { get } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import {
   Col,
@@ -15,7 +16,6 @@ import css from './DocumentCard.css';
 
 export default class DocumentCard extends React.Component {
   static propTypes = {
-    atType: PropTypes.string,
     location: PropTypes.string,
     name: PropTypes.string.isRequired,
     note: PropTypes.string,
@@ -52,7 +52,7 @@ export default class DocumentCard extends React.Component {
 
   render() {
     const { location, name, note, url } = this.props;
-    const category = this.props.atType;
+    const category = get(this.props, ['atType', 'label']);
 
     const contentData = [];
     if (location) contentData.push({ location });
--- a/lib/DocumentCard/DocumentCard.js
+++ b/lib/DocumentCard/DocumentCard.js
@@ -64,7 +64,7 @@ export default class DocumentCard extends React.Component {
           {name}
         </Headline>
         <Row>
-          <Col xs={8}>
+          <Col xs={category ? 8 : 12}>
             {note ?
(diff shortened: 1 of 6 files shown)


lib/DocumentCard/DocumentCard.js
```diff
@@ -59,13 +63,26 @@ export default class DocumentCard extends React.Component {
...
+          <Col xs={8}>
+            {note ?
+              <KeyValue label={<FormattedMessage id="stripes-erm-components.doc.note" />}>
+                <div data-test-doc-note>{note}</div>
+              </KeyValue>
+              :
+              null
+            }
+          </Col>
+          <Col>
```
doytch: This should have an `xs={category ? 4 : 0}` for completeness. Generally, always have a size when using `Col`.

lib/DocumentCard/DocumentCard.js
```diff
@@ -59,13 +63,26 @@ export default class DocumentCard extends React.Component {
...
         </Headline>
-        {note ?
-          <KeyValue label={<FormattedMessage id="stripes-erm-components.doc.note" />}>
-            <div data-test-doc-note>{note}</div>
-          </KeyValue>
-          :
-          null
-        }
+        <Row>
+          <Col xs={8}>
```
doytch: Let's let the `Col` be the full width of 12 if there's no category. So, `xs={category ? 8 : 12}`

lib/DocumentsFieldArray/DocumentsFieldArray.js
```diff
@@ -60,8 +62,31 @@ class DocumentsFieldArray extends React.Component {
...
+          <FormattedMessage id="ui-licenses.placeholder.documentCategory">
+            {placeholder => (
+              <Field
+                component={Select}
+                dataOptions={documentCategories}
+                id={`${name}-category-${i}`}
+                label={<FormattedMessage id="stripes-erm-components.doc.category" />}
+                name={`${name}[${i}].atType`}
+                placeholder={placeholder}
+                value="no"
```
doytch: There shouldn't be a hardcoded `value` here, because that forces the `Field` to _always_ have the value of `"no"` as opposed to whatever the user selected. What was the goal with this prop?

lib/DocumentsFieldArray/DocumentsFieldArray.js
```diff
@@ -60,8 +62,30 @@ class DocumentsFieldArray extends React.Component {
     !value ? <FormattedMessage id="stripes-core.label.missingRequiredField" /> : undefined
   )
 
+  renderCategory = (i) => {
+    const { documentCategories, name } = this.props;
+    return (
+      <Row>
+        <Col xs={11}>
+          <FormattedMessage id="ui-licenses.placeholder.documentCategory">
```
doytch: This shouldn't depend on a `ui-licenses` translation string.

PR #40, opened 2019-04-11
New Tags erm component, refs ERM-173

aditya-matukumalli:
- Has withTags HOC to facilitate tags helper app
- Performs all the tags related CRUD operations
- added translations
- Readme

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,8 @@ export { default as LicenseEndDate } from './lib/LicenseEndDate';
 export { default as LicenseTermsList } from './lib/LicenseTermsList';
 export { default as OrganizationSelection } from './lib/OrganizationSelection';
 export { default as Spinner } from './lib/Spinner';
-
+export { default as withTags } from './lib/Tags/withTags';
+export { default as Tags } from './lib/Tags';
 // HOCs
 export { default as withKiwtFieldArray } from './lib/withKiwtFieldArray/withKiwtFieldArray';
 
--- a/lib/Tags/Tags.js
+++ b/lib/Tags/Tags.js
@@ -1,0 +1,141 @@
+import { get, uniqBy, sortBy, difference, cloneDeep } from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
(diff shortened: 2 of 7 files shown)


doytch:
Wait, do we really need _all_ of these components? I thought the only one we needed to port was the `Tags` container component.

aditya-matukumalli:
@doytch I thought the same initially, but looks like we need atleast the _Tags_ and _TagsForm_ components. We can get rid of the _withTags_ component ( ported it accidentally ). We need the _TagsForm_ as the _onChange_ handlers  behavior is different here due to the data object shape that we get back.

aditya-matukumalli:
@doytch NVM. I modeled the necessary data to be passed to _TagsForm_ in the _Tags_ component and passed it down. Got rid of both _TagsForm_ and the _withStripes_ components.

lib/Tags/Tags.js
```diff
@@ -0,0 +1,142 @@
...
+import { Callout, Pane } from '@folio/stripes/components';
+import { stripesConnect } from '@folio/stripes/core';
+
+import { TagsForm } from '@folio/stripes/smart-components';
+
+class Tags extends React.Component {
+  static manifest = Object.freeze({
+    tags: {
+      type: 'okapi',
+      path: 'tags?limit=100',
```
doytch: 100? This seems low, but I guess you c/ped this limit from the SSC implementation, right?

index.js
```diff
@@ -7,7 +7,7 @@ export { default as LicenseEndDate } from './lib/LicenseEndDate';
 export { default as LicenseTermsList } from './lib/LicenseTermsList';
 export { default as OrganizationSelection } from './lib/OrganizationSelection';
 export { default as Spinner } from './lib/Spinner';
-
+export { default as Tags } from './lib/Tags';
```
doytch: Nit picking, but leave an empty line between the "sections" of exports.

lib/Tags/Tags.js
aditya-matukumalli: Yes

PR #51, opened 2019-04-26
Erm setup tests

aditya-matukumalli:


lib/LicenseTermsList/tests/interactor.js
```diff
@@ -0,0 +1,22 @@
...
+  isPresent,
+  count,
+  collection
+} from '@bigtest/interactor';
+
+@interactor class LicenseTermInteractor {
+  isLabelPresent = isPresent('[data-test-term-label]');
+  isNamePresent = isPresent('[data-test-term-name]');
+  label = text('[data-test-term-label]');
+  name = text('[data-test-term-name]');
```
doytch: So this should be changed if reverting that change to the dataset props. More accurately and sensibly now, it'd have a `value` prop for the interactor instead.

package.json
```diff
@@ -40,6 +45,6 @@
     "react": "*"
   },
   "stripes": {
-    "type": "components"
+    "type": ""
```
doytch: Are you sure about this? I've talked about this setting with Matt in the past.

package.json
```diff
@@ -30,6 +31,10 @@
     "react-dom": "^16.6.3"
   },
   "dependencies": {
+    "@bigtest/convergence": "^1.1.1",
+    "@bigtest/mirage": "0.0.1",
+    "@folio/stripes-components": "^5.2.1000775",
```
doytch: Noooo, why are we depending on CI versions of stripes-components and stripes-core here?

lib/CreateOrganizationModal/CreateOrganizationModalContainer.js
```diff
@@ -40,12 +41,12 @@ export default class CreateOrganizationModal extends React.Component {
...
       .GET({ params: { match: 'name', term: this.state.name } })
       .then(results => {
-        if (!results || results.length) {
+        if (results && results.length) {
           this.setState({ error: <FormattedMessage id="stripes-erm-components.createOrg.alreadyExists" /> });
         } else {
           organizations
             .POST({ name: this.state.name })
-            .then(() => { this.props.onClose(); })
+            .then(() => this.props.onClose)
```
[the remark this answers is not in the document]
aditya-matukumalli: Ah, was testing something else and was supposed to add _onClose()_ but added _onClose_.

lib/LicenseCard/tests/LicenseCard-test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    it('renders the licensor', () => {
+      expect(licensecard.isLicensorPresent).to.be.true;
+    });
+
+    it('sets licensor to license card licensor', () => {
+      const { orgs = [] } = license;
+      const licensor = orgs.find(o => get(o, ['role', 'value']) === 'licensor');
+      const licensorName = get(licensor, ['org', 'name']);
```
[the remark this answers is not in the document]
aditya-matukumalli: I didnt. Good to know.

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -36,11 +36,11 @@ export default class LicenseTermsList extends React.Component {
       <Layout className="padding-top-gutter" key={index}>
         <Row>
           <Col xs={12} md={4}>
-            {term.description ? <InfoPopover content={term.description} /> : null}
+            {term.description ? <InfoPopover data-test-info-popover content={term.description} /> : null}
             <strong data-test-term-label={term.name}>{term.label}</strong>
           </Col>
           <Col xs={12} md={8}>
-            <span data-test-term-value={term.name}>{value}</span>
+            <span data-test-term-name={term.name}>{value}</span>
```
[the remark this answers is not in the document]
aditya-matukumalli: Changed it back to _data-test-term-value_. We use this hook only to see if its present in the dom and eventually grab the _value_ there. We don't do anything with the _term.name_.

package.json
doytch: These should also be dev dependencies, right?

package.json
aditya-matukumalli: Yup! touched base with Matt on this. He pointed me to https://issues.folio.org/browse/STCLI-107 which would be a fix coming soon. So, the _"type": "components"_ instructs the CLI to skip a platform build when running unit tests. As not all erm-components exactly "dumb" components and we  need the platform to run tests, we shouldn't be having the _"type": "components"_ in the package.

package.json
aditya-matukumalli: I have absolutely no idea how that made it through. My bad. Must be with some copy over when resolving lint issues.

package.json
doytch: Awesome! 👏

PR #59, opened 2019-05-10
Erm organizations field array

aditya-matukumalli:


lib/OrganizationsFieldArray/OrganizationPicker.js
```diff
@@ -0,0 +1,57 @@
...
+
+    return (
+      <FormattedMessage id="stripes-erm-components.organizations.clickSearchButton">
+        {placeholder => (
+          <TextField
+            error={error}
+            endControl={(
+              <Pluggable
+                aria-haspopup="true"
+                dataKey="organization"
```
doytch: If you reuse this `dataKey`, does the plugin reset its state when opened multiple times? Ie, when you open modal, set some filters and search query, pick an org, add another org, and open the modal again, are the filters/query default or are they the ones you set previously.

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -0,0 +1,161 @@
...
+    addContactBtnLabel: <FormattedMessage id="stripes-erm-components.organizations.addOrganization" />,
+    isEmptyMessage: <FormattedMessage id="stripes-erm-components.organizations.noOrganizations" />,
+    organizationPickerComponent: OrganizationPicker,
+  }
+
+  state = {
+    organizations: {},
+  };
+
+  static getDerivedStateFromProps(props, state) {
```
doytch: Do you need to be merging everything into `state` for this component? I don't really see `state` being used anywhere, though I may be missing something. 

(Internal contacts does it mainly because we need to merge the previously-set contacts with complete user info that's fetched from the `/users` endpoint.)

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -0,0 +1,161 @@
...
+  validateDuplicateEntries = /* istanbul ignore next */ (value, allValues) => {
+    const orgs = allValues[this.props.name] || [];
+
+    const organizations = orgs.filter(organization => {
+      const { org } = organization;
+      return org ? org.name === value : [];
+    });
+    if (organizations.length < 2) return undefined;
+
+    const rolesForThisorg = orgs.map(org => org.role);
```
doytch: Capitalization 😆

lib/OrganizationsFieldArray/OrganizationPicker.js
```diff
@@ -0,0 +1,57 @@
...
+                disableRecordCreation
+                marginTop0
+                searchButtonStyle="fieldControl"
+                selectVendor={org => {
```
doytch: They didn't change the callback name when moving from `find-vendor` to `find-organization`? 😬 if so.

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -0,0 +1,161 @@
...
+
+  addOrganization = /* istanbul ignore next */ (org, index) => {
+    const { name } = this.props;
+
+    this.setState(prevState => ({
+      organizations: {
+        ...prevState.organizations,
+        [org.id]: org,
+      }
+    }), () => this.props.change(`${name}[${index}].org.orgsUuid`, org.id)); // creates a new field in redux-form for orgsUuid
```
doytch: I'm actually not in love with this approach anymore. I think another way of doing this, rather than having to carry `change` all the way down the component tree, would be to implement an `OrganizationField` component that basically renders/manages the components/object for the `${name}[${index}]`.

So that component's value would be something like `{ role: 'abc', org: { name: 'def', orgsUuid: 'ghi' } }`, and it can render controlled components directly (as opposed to `Field`s) and be able to easily manage all the callbacks/rendering.

I think that would actually end up being a lot cleaner in this component and wherever it's used.

lib/OrganizationsFieldArray/tests/OrganizationsFieldArray-test.js
```diff
@@ -0,0 +1,133 @@
...
+import { expect } from 'chai';
+import { FieldArray } from 'redux-form';
+import { TextField } from '@folio/stripes/components';
+
+import { mountWithContext } from '../../../tests/helpers';
+import TestForm from '../../../tests/TestForm';
+import OrganizationsFieldArray from '../OrganizationsFieldArray';
+import OrganizationsFieldArrayInteractor from './interactor';
+
+const addContactBtnLabel = 'Add contact';
```
doytch: Bunch of references to contacts in here.

lib/OrganizationsFieldArray/tests/interactor.js
```diff
@@ -0,0 +1,29 @@
...
+  hasOrg = isPresent('[data-test-org-name]');
+  hasRole = isPresent('[data-test-org-role]');
+  orgValue = value('[data-test-org-name]');
+  roleValue = value('[data-test-org-role]');
+  orgText = text('[data-test-org-name]');
+  roleText = text('[data-test-org-role]');
+  setOrg = fillable('[data-test-org-name]');
+  setRole = selectable('[data-test-org-role]');
+  clickAddButton = clickable('[data-test-orgfa-add-button]');
+  addContactBtnLabel = text('[data-test-orgfa-add-button]');
```
doytch: Contacts...

lib/OrganizationsFieldArray/OrganizationPicker.js
aditya-matukumalli: Yes, it is resetting the defaults.

lib/OrganizationsFieldArray/OrganizationPicker.js
aditya-matukumalli: No, they didn't :p

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
aditya-matukumalli: Good catch. We don't _state_ at all in this file

PR #61, opened 2019-05-14
Pass id to org Picker, refs ERM-220

aditya-matukumalli:


lib/OrganizationsFieldArray/OrganizationPicker.js
```diff
@@ -10,6 +10,7 @@ export default class OrganizationPicker extends React.Component {
     meta: PropTypes.shape({
       error: PropTypes.node,
     }),
+    inputId: PropTypes.string,
```
doytch: Can we call this `id` for consistency with other components?

PR #62, opened 2019-05-15
ERM-230: Added Card Component

doytch:
Added a basic `Card` component to let us render the card-based layouts we're about to start using often in ERM apps. This was inspired by the address edit UI in the Users app and picked up by our UX designer after consultation with Filip.

**Basic Usage**
test-card

**In The Wild**
card

Eventually, I believe this is a candidate for being in @folio-org/stripes itself.

### DIFF
--- a/lib/Card/Card.js
+++ b/lib/Card/Card.js
@@ -21,7 +21,6 @@ export default class Card extends React.Component {
         <div
           className={css.header}
           data-test-card-header
-          start="xs"
         >
           {header}
         </div>
--- a/lib/Card/Card.js
+++ b/lib/Card/Card.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import css from './Card.css';
+
+export default class Card extends React.Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    header: PropTypes.node.isRequired,
+  }
+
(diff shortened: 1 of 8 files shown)


mkuklis:
It looks like a very nice component which perhaps should be moved to stripes-components?

doytch:
I agree, @mkuklis! This was my way of asking if others feel the same way. 😆

lib/Card/Card.js
```diff
@@ -0,0 +1,37 @@
...
+      header,
+      ...rest
+    } = this.props;
+
+    return (
+      <div className={css.card} {...rest}>
+        <div
+          className={css.header}
+          data-test-card-header
+          start="xs"
```
rasmuswoelk: What does this prop do?

doytch: Nothing, it's an artifact.

doytch:
I'm gonna go ahead and merge this here now since we have a need for it now for our sprint ending tomorrow. The versioning issues (depending on unreleased features) would prevent us from being able to use this in that short a time frame otherwise.

I _will_ open a PR against stripes-components with this soon (probably next week) with an eye to switching to that instead.

rasmuswoelk:
@doytch Cool! I think it will be a good addition for stripes-components. I'll review your PR when it's up 👍 Thanks!

PR #66, opened 2019-05-23
ERM-241: Update the UI for license terms (to take account of the new …

CalamityC:
…note field)

change the layout of terms using the Card component

### DIFF
--- a/lib/LicenseTermsList/LicenseTermsList.js
+++ b/lib/LicenseTermsList/LicenseTermsList.js
@@ -6,7 +6,6 @@ import {
   Col,
   InfoPopover,
   KeyValue,
-  Layout,
   Row
 } from '@folio/stripes/components';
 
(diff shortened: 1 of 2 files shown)


lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -35,24 +38,32 @@ export default class LicenseTermsList extends React.Component {
...
-            <strong data-test-term-label={term.name}>{term.label}</strong>
-          </Col>
-          <Col xs={12} md={8}>
-            <span data-test-term-value={term.name}>{value}</span>
-          </Col>
-        </Row>
-        { note ? (
+      <Layout key={index}>
+        <Card
+          id="term-card"
```
doytch: This will create multiple items with the same `id` and `id`s should be unique across an entire page. Consider appending the `index` to the `id` like:
```suggestion
          id={`term-card-${index}`}
```

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -35,24 +38,32 @@ export default class LicenseTermsList extends React.Component {
...
-          <Col xs={12} md={4}>
-            {term.description ? <InfoPopover data-test-info-popover content={term.description} /> : null}
-            <strong data-test-term-label={term.name}>{term.label}</strong>
-          </Col>
-          <Col xs={12} md={8}>
-            <span data-test-term-value={term.name}>{value}</span>
-          </Col>
-        </Row>
-        { note ? (
+      <Layout key={index}>
```
doytch: I don't think this `Layout` element is necessary here. `Layout` only does something if you pass it a `className` prop.

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -35,24 +38,32 @@ export default class LicenseTermsList extends React.Component {
...
+        <Card
+          id="term-card"
+          header={
+            <Col xs={12}>
```
doytch: I think a `span` or `div` is fine here rather than a `Col`. 

`Col`s should only be used inside of a `Row` because of the way they assume the other component applied margins.

CalamityC:
thanks for your comments, i made all the changes and attached a screenshot
terms

PR #77, opened 2019-06-27
Erm 230 editorgs

aditya-matukumalli:
- Added new _OrganizationsFieldArray_ component that renders an __OrganizationCard__ component along with organization role selector.
- Added new _OrganizationCard_ component used to render interfaces/orgName/default card layouts conditionally.
- Added new _onMarkforDeletion_ prop to WithKiwtFieldArray that marks a field for deletion.


lib/OrganizationCard/OrganizationCard.js
```diff
@@ -0,0 +1,115 @@
...
+            <React.Fragment>
+              <InterfacesCard name={name} index={index} organization={organization} {...this.props} />
+              {this.renderRole(name, index)}
+            </React.Fragment>
+          </Card>
+        );
+      });
+
+      return (
+        <React.Fragment>
```
doytch: Why return this in a Fragment?

lib/OrganizationCard/OrganizationCard.js
```diff
@@ -0,0 +1,115 @@
...
+            headerStart={<strong>{`Organization ${index + 1}`}</strong>}
+            headerEnd={<IconButton
+              data-test-org-delete-button
+              id={`${name}-delete-${index}`}
+              onClick={() => onDeleteField(index, organization)}
+              icon="trash"
+            />}
+          >
+            <React.Fragment>
+              <InterfacesCard name={name} index={index} organization={organization} {...this.props} />
```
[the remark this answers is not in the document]
doytch: Also, don't pass down `this.props` like this carelessly. Why would an InterfacesCard need the `onAddField` or `onDeleteField` props?

package.json
```diff
@@ -43,6 +43,7 @@
     "prop-types": "^15.6.0",
     "react-dropzone": "^10.0.0",
     "react-intl": "^2.4.0",
+    "react-router-dom": "^5.0.1",
```
doytch: Why?

translations/stripes-erm-components/en.json
```diff
@@ -35,7 +35,10 @@
   "organizations.role": "Role",
   "organizations.addOrganization": "Add organization",
   "organizations.noOrganizations": "No organizations have been added.",
-
+  "organizations.noOrganization": "No organizations added",
+  "organizations.addOrganizationToStart": "Add an organization to get started",
+  "organizations.addOrganizationToAgreement": "Add organization to agreement",
+  "organizations.unlinkOrganization": "Unlink organization",
```
doytch: Please be mindful of whitespace. There's a reason there's an empty line between each section.

lib/OrganizationCard/InterfacesCard.js
```diff
@@ -0,0 +1,182 @@
...
+              size="small"
+            />
+            {' '}
+            <strong>Organization</strong>
+          </span>
+        }
+        headerEnd={this.renderOrganizationPicker(index, name)}
+        roundedBorder
+      >
+        <React.Fragment>
```
doytch: Why is this a Fragment? Don't overuse Fragments please.

lib/OrganizationCard/InterfacesCard.js
```diff
@@ -0,0 +1,182 @@
...
+import {
+  Button,
+  Card,
+  Layout,
+  MultiColumnList,
+  Icon,
+} from '@folio/stripes/components';
+import OrganizationPicker from './OrganizationPicker';
+
+export default class InterfacesCard extends React.Component {
```
doytch: So this isn't really just an InterfacesCard is it? It renders the entire Organization card...and may list no organizations whatsoever. Can we clean up the way all these components relate to one another? It's super confusing.

doytch: Also, this component is too big. It renders two different kinds of cards that do different things because one of them is in a Field and one isn't.

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -1,144 +1,72 @@
...
-            data-test-org-delete-button
-            icon="trash"
-            id={`${name}-delete-${index}`}
-            onClick={() => onDeleteField(index, org)}
+    renderOrganizations = (items) => {
+      const orgs = items.map((organization, index) => {
+        return (
+          <OrganizationCard
+            organization={organization}
+            index={index}
```
doytch: TODO: Issue?

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -1,144 +1,72 @@
...
+            {...this.props}
```
doytch: Is this necessary? Pass down only what's needed.

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -1,144 +1,72 @@
...
-      </React.Fragment>
-    );
-  }
+        );
+      });
 
-  render() {
-    const { items, onAddField } = this.props;
+      return (
+        <React.Fragment>
```
doytch: Why is this in a Fragment?

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,60 @@
...
+      organizationPickerComponent,
+      roles,
+    } = this.props;
+
+    return (
+      <Card
+        data-test-organizations-org
+        hasMargin
+        key={index}
+        headerStart={<strong>{`Organization ${index + 1}`}</strong>}
```
doytch: i18n

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Card,
+  IconButton,
+} from '@folio/stripes/components';
+import OrgInfo from './OrgInfo';
+
+export default class OrganizationCard extends React.Component {
```
doytch: It might make sense to call this something like `EditOrganizationCard`, since we're gonna have a card within it that should be the one that's _actually_ called `OrganizationCard`.

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,60 @@
...
+  static propTypes = {
+    name: PropTypes.string,
+    index: PropTypes.string,
+    organization: PropTypes.object,
+    onDeleteField: PropTypes.func,
+    onReplaceField: PropTypes.func,
+    onAddField: PropTypes.func,
+    organizationPickerComponent: PropTypes.elementType,
```
doytch: Is it possible to remove this whole `organizationPickerComponent` thing? Would it simplify this whole set of components if we removed it since it looks like we're passing it down several levels? It was only there to get higher coverage stats and it didn't really work out.

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+  Button,
+  Col,
+  Card,
+  Layout,
+  Label,
+  Select,
+} from '@folio/stripes/components';
+import { required } from '../validators';
+
+export default class OrgInfo extends React.Component {
```
doytch: Could we change this component to:
1. Not do the decision making of whether to render the org info vs org picker, moving that logic up to the `EditOrganizationCard`. So of its current responsibilities, remove everything that isn't rendering the org name and role.
1. Support rendering of interfaces if they are passed into it as a prop.
1. Rename this to `OrganizationCard`

?

If so, could we then use this component in _both_ the `EditOrganizationCard` as well as where we're currently rendering out the organizations in `ui-agreements/src/components/AgreementSections/Organizations.js`? That would be nice.

doytch: In any case, the decision making that I say in bullet 1 should probably be moved up to that hypothetical `EditOrganizationCard` (what's currently just called `OrganizationCard`). This design results in a lot of props passing when it wouldn't be strictly necessary.

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+            />
+          )}
+        </FormattedMessage>
+      </Col>);
+  }
+
+  renderDefault = () => {
+    const { index, name } = this.props;
+    return (
+      <Field
```
doytch: Why is this a Field rendering a Card? Card has no handling for any of the props that redux-form will give it.

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+    );
+  }
+
+  onAddOrg = (organization, index) => {
+    const { name, id } = organization;
+    const org = { name, orgsUuid: id };
+    this.props.onReplaceField(index, { org });
+  }
+
+  unlinkOrg = (index, orgId) => {
```
doytch: This definitely needs comments explaining what's happening here.

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+    this.props.onReplaceField(index, { org: {} });
+  }
+
+  renderOrganizationName = (orgName) => {
+    const { index, name } = this.props;
+    return (
+      <span>
+        <AppIcon
+          app="organizations"
+          appIconKey="organizations"
```
doytch: Is `appIconKey` necessary here?

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+
+  renderOrgInfo = () => {
+    const { index, organization, name } = this.props;
+    const { org } = organization;
+    return (
+      <Card
+        cardStyle="positive"
+        hasMargin
+        headerStart={
+          <strong>Organization</strong>
```
doytch: i18n. Can you do a pass through this entire PR checking for untranslated strings?

lib/OrganizationsFieldArray/OrgInfo.js
```diff
@@ -0,0 +1,162 @@
...
+        data-test-org-name
+        id={`${name}-nameOrg-${index}`}
+        name={`${name}[${index}].org`}
+        cardStyle="negative"
+        hasMargin
+        headerStart={
+          <span>
+            <AppIcon
+              app="organizations"
+              appIconKey="organizations"
```
doytch: Is `appIconKey` necessary here?

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -1,144 +1,83 @@
...
+      <Layout className="padding-bottom-gutter" data-test-org-empty-message>
+        {this.props.isEmptyMessage}
+      </Layout>
     );
-  }
 
-  render() {
-    const { items, onAddField } = this.props;
+    renderOrganizations = (items) => {
+      const { onMarkforDeletion, name, roles, organizationPickerComponent, onDeleteField, onReplaceField } = this.props;
```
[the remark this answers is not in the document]
doytch: Also, we've been trying to always alphabetize destructuring statements and prop lists. There's a few examples of that throughout this PR. It just makes it a lot easier to skim the code later.

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -1,144 +1,83 @@
...
+      const orgs = items.map((organization, index) => {
+        return (
+          <EditOrganizationCard
+            name={name}
+            organization={organization}
+            index={index}
+            items={items}
```
doytch: Can you please be careful and conscious about the prop passing you're doing? I don't see why `items` needs to be passed in, and it's not declared in the prop types for `EditOrganizationCard`. This kind of stuff has come up a lot of times during this PR and it's tough to constantly scroll back and forth through Github trying to figure whether I'm missing something or not.

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -0,0 +1,229 @@
...
+      validate,
+      renderDefault
+    };
+  };
+
+  renderOrganizationCard = () => (
+    <Field
+      data-test-org-name
+      component={OrganizationCard}
+      {...this.getOrganizationCardProps()}
```
doytch: Is it really necessary to define this in a separate function? I don't see it being used anywhere else...

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -0,0 +1,229 @@
...
+        onClick={() => this.unlinkOrg()}
+      >
+        <FormattedMessage id="stripes-erm-components.organizations.unlinkOrganization" />
+      </Button>
+    );
+  }
+
+  renderHeaderStart = (isOrgPresent) => {
+    return (
+      isOrgPresent ?
```
doytch: Could we just always render the same thing here, with the `AppIcon`?

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -0,0 +1,229 @@
...
+            </strong>
+          </span>
+        )
+    );
+  };
+
+  renderHeaderEnd = (isOrgPresent, id) => {
+    const { index, name } = this.props;
+    return isOrgPresent ?
+      this.renderOrgInfoHeaderEnd(id) :
```
doytch: Seems like these two functions don't really need to be defined in separate functions. It just leads to inconsistencies like passing `id` to `renderOrgInfoHeaderEnd` when it doesn't use that param.

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -0,0 +1,229 @@
...
+          <FormattedMessage id="stripes-erm-components.organizations.noOrganization" />
+        </strong>
+      </Layout>
+      <Layout className="textCentered">
+        <FormattedMessage id="stripes-erm-components.organizations.addOrganizationToStart" />
+      </Layout>
+    </div>
+  );
+
+  onAddOrg = (organization, index) => {
```
doytch: We've been trying to keep a pattern of callback handlers being called `handleFoo` and being passed as `onFoo`. So this would be called `handleAddOrg` and the below would be called `handleUnlinkOrg`. It makes it easier to track where something is defined rather than where it's just being used or passed through.

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,134 @@
...
+    id: PropTypes.string,
+    interfaces: PropTypes.object,
+    meta: PropTypes.shape({
+      error: PropTypes.node,
+    }),
+    org: PropTypes.object,
+    renderDefault: PropTypes.func,
+  };
+
+  renderError = (error) => (
```
doytch: This function isn't called anywhere.

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,134 @@
...
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    headerEnd: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    id: PropTypes.string,
+    interfaces: PropTypes.object,
+    meta: PropTypes.shape({
```
doytch: With this and the way you're doing the error handling, you're expressing the dependency that `Card` should a redux form field. This isn't ideal and we don't even always use it like that ourselves, like when we're on the view agreement/license pane.

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,134 @@
...
+              {item.name}
+              <a href={item.uri}>
+                <Icon icon="external-link" iconPosition="end" />
+              </a>
+            </span>
+          ),
+          notes: item => get(item, 'notes'),
+          username: item => get(item, 'username'),
+          password: item => get(item, 'password'),
+          type: item => get(item, 'type'),
```
doytch: [`type` is an array](https://s3.amazonaws.com/foliodocs/api/mod-organizations-storage/interface.html#organizations_storage_interfaces__id__get)

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,134 @@
...
+          </strong>
+        </Layout>
+      </div>
+    );
+  }
+
+  render() {
+    const {
+      org,
+      renderDefault,
```
doytch: Can we call this `renderEmpty`? "Default" is pretty ambiguous. Ie, default compared to what?

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -0,0 +1,229 @@
...
+    onReplaceField: PropTypes.func,
+    roles: PropTypes.arrayOf(PropTypes.object),
+  };
+
+  renderOrgInfoHeaderEnd = () => {
+    const { name, index } = this.props;
+    return (
+      <Button
+        data-test-unlink-button
+        buttonStyle="danger"
```
doytch: This button is misaligned vertically because of bottom margin. [Apply `marginBottom0` to it.](https://github.com/folio-org/stripes-components/tree/master/lib/Button#props)

Same thing with the "Add Organization" button when an org hasn't been selected.

lib/OrganizationsFieldArray/OrganizationCard.js
```diff
@@ -0,0 +1,134 @@
...
+  Card,
+  Icon,
+  MultiColumnList,
+  Layout,
+} from '@folio/stripes/components';
+import css from './OrganizationCard.css';
+
+export default class OrganizationCard extends React.Component {
+  static propTypes = {
+    cardStyle: PropTypes.string,
```
doytch: I get why you're passing `cardStyle` in here but the whole point of this card is so that we don't have to remember what particular props to set on `Card` to have the right styling. At the very least, this should probably have a default prop value so that common usage of rendering an organization doesn't require `cardStyle` to be passed in.

It's part of what makes me unsure whether it's actually appropriate for this component to be used when rendering both a selected and unselected organization. I kind of feel like this should only be used for a _selected_ organization. We can talk about this if you want.

lib/OrganizationsFieldArray/OrganizationCard.js
doytch: I guess this also loops back to the above comment talking about whether "this component [should] be used when rendering both a selected and unselected organization". 

If it was only used when rendering a _selected_ organization, then you wouldn't need handle displaying errors in this component since the "role" dropdown could always be responsible for it. Then you wouldn't need this dependency on `meta` supplied by redux-form.

lib/OrganizationsFieldArray/EditOrganizationCard.js
aditya-matukumalli: Done

lib/OrganizationsFieldArray/EditOrganizationCard.js
aditya-matukumalli: _Add Organization_ Button is rendered here https://github.com/folio-org/ui-plugin-find-organization/blob/master/OrganizationSearch/OrganizationSearch.js#L51 and looks like it doens't have a capability to render a _marginBottom0_ style prop. Might have to raise a PR against ui-plugin0find-org and add it.

aditya-matukumalli: Raised a PR for this against the plugin

aditya-matukumalli:
@doytch 
- Added the documentation readme files
- Moved _ViewOrganizationCard_ into its own file
- Added Tests.

I wasn't able to make use of the Plugin in the tests somehow. So just used the TextField as we did previously in the tests as is in there for the _OrganiztionsFieldArray._ So, It Didn't make sense to write tests for _EditOrganizationCard_ separately. Added in few tests for the _ViewOrganizationCard_ component.

lib/ViewOrganizationCard/tests/ViewOrganizationCard-test.js
```diff
@@ -0,0 +1,72 @@
...
+
+  it('renders the card header', () => {
+    expect(interactor.hasHeaderStart).to.be.true;
+  });
+
+  it('renders the card body', () => {
+    expect(interactor.hasCardBody).to.be.true;
+  });
+
+  it('renders a Multi column list', () => {
```
doytch: Can we check the validity of this list? Eg, some checks for the contents matching the interfaces we're passing in?

lib/ViewOrganizationCard/ViewOrganizationCard.js
```diff
@@ -0,0 +1,110 @@
...
+    const { id } = this.props;
+    return (
+      <span>
+        <AppIcon
+          app="organizations"
+          size="small"
+        />
+        &nbsp;
+        <strong id={id} data-test-org-name>
+          {`${orgName}`}
```
doytch: This can just be `{orgName}`

lib/OrganizationsFieldArray/SelectOrganizationCard.js
```diff
@@ -0,0 +1,57 @@
...
+      <div>
+        <Card
+          cardStyle="negative"
+          data-test-organization-card
+          hasMargin
+          headerStart={headerStart}
+          headerEnd={headerEnd}
+          roundedBorder
+        >
+          { error ? this.renderError(error) : (renderEmpty && renderEmpty())}
```
doytch: It's up to you, but would it make more sense to just use `this.props.children` instead of `renderEmpty` for this kind of thing? You could always render `this.props.children` to make sure any instructions in there are visible even when there's an error as well?

lib/ViewOrganizationCard/tests/ViewOrganizationCard-test.js
aditya-matukumalli: Yup, will add that in

lib/OrganizationsFieldArray/SelectOrganizationCard.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+import {
+  Card,
+  Layout
+} from '@folio/stripes/components';
+import css from './OrganizationCard.css';
+
+export default class SelectOrganizationCard extends React.Component {
+  static propTypes = {
+    children: PropTypes.func,
```
doytch:
```suggestion
    children: PropTypes.node,
```

PR #79, opened 2019-07-03
Show spinner when File upload is in progress

aditya-matukumalli:
- Renders a _spinner_ (erm-component) in place of the _uploadText_ when large file uploads/uploads on sub-par connections are in progress to let the user know that the request is still in flight.
- Spinner hides itself once the request resolves to a success or an error


lib/DocumentsFieldArray/FileUploaderField.js
```diff
@@ -45,6 +47,7 @@ class FileUploaderField extends React.Component {
       isDropZoneActive: false,
     });
 
+    this.setState({ uploadInProgress: true });
```
doytch: You can just loop this into the `setState` above, no?

lib/FileUploader/FileUploader.js
```diff
@@ -70,7 +73,7 @@ const FileUploader = props => {
             className={`${css.uploadTitle} ${isDropZoneActive ? css.activeUploadTitle : ''}`}
             data-test-title
           >
-            {title}
+            {uploadInProgress ? <Spinner /> : title}
```
doytch: We should probably also have a little "Uploading" label to drive home the point of what's happening. And remove the `Button` we render below if we're showing the `Spinner`.

May need to convert this component to a class-based one to accomplish that the cleanest way possible, but this render function is getting big enough already so it could be for the best.

lib/DocumentsFieldArray/FileUploaderField.js
aditya-matukumalli: Oh yeah, I missed that block of code somehow. Will add it in there.

PR #80, opened 2019-07-08
Display appropriate error on file uploads, refs ERM-277

aditya-matukumalli:
Handles the case where the error from the response may not be _json_ (of the type _application/json_)and in that case we just display a generalized error message

### DIFF
--- a/lib/DocumentsFieldArray/FileUploaderField.js
+++ b/lib/DocumentsFieldArray/FileUploaderField.js
@@ -41,11 +41,11 @@ class FileUploaderField extends React.Component {
 
 
   processError(resp, intl) {
-    const contentType = resp.headers.get('Content-Type') || '';
+    const contentType = resp.headers ? resp.headers.get('Content-Type') : '';
     if (contentType.startsWith('application/json')) {
       throw new Error(`${resp.message} (${resp.error})`);
     } else {
-      throw new Error(intl.formatMessage({ id: 'ui-agreements.errors.uploadError' }));
+      throw new Error(intl.formatMessage({ id: 'stripes-erm-components.errors.uploadError' }));
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/DocumentsFieldArray/FileUploaderField.js
```diff
@@ -39,7 +39,17 @@ class FileUploaderField extends React.Component {
...
   }
 
-  handleDrop = (acceptedFiles) => {
+
+  processError(resp, intl) {
+    const contentType = resp.headers.get('Content-Type') || '';
+    if (contentType.startsWith('application/json')) {
+      throw new Error(`${resp.message} (${resp.error})`);
+    } else {
+      throw new Error(intl.formatMessage({ id: 'ui-agreements.errors.uploadError' }));
```
doytch: Only use this module's translation keys within it.

lib/DocumentsFieldArray/FileUploaderField.js
```diff
@@ -50,12 +60,14 @@ class FileUploaderField extends React.Component {
 
     this.props.onUploadFile(acceptedFiles[0])
       .then(response => {
-        if (response.error) {
-          throw new Error(`${response.message} (${response.error})`);
+        if (response.ok) {
+          response.json().then(resp => {
```
doytch: Nit-picking (since this'll need a respin anyway 😆) but this could be a tad cleaner if you called the variable `resp` instead of `file`.

lib/DocumentsFieldArray/FileUploaderField.js
```diff
@@ -74,18 +86,22 @@ class FileUploaderField extends React.Component {
...
-        isDropZoneActive={this.state.isDropZoneActive}
-        onDelete={this.handleDelete}
-        onDownloadFile={this.props.onDownloadFile}
-        onDragEnter={() => this.setState({ isDropZoneActive: true })}
-        onDragLeave={() => this.setState({ isDropZoneActive: false })}
-        onDrop={this.handleDrop}
-        uploadInProgress={this.state.uploadInProgress}
-        {...pickBy(this.props, (_, key) => key.startsWith('data-test-'))}
-      />
+      <IntlConsumer>
```
doytch: Woof. This looks _rough_, but I don't really see a better way of writing it. Seems like the kind of thing we should refactor to use `injectIntl` when we get onto `react-intl` v3.0. Can you put a comment above this `return` saying that so we don't forget? Something like:

```/* TODO: Refactor this component to use `injectIntl` when Folio starts using react-intl 3.0 */```

PR #98, opened 2019-07-31
ERM-355: Manage public notes on license term values

CalamityC:
change terms accordion in license preview pane
* add Public note
* change Note to Internal Note
* add Visibility


lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -19,9 +19,18 @@ export default class LicenseTermsList extends React.Component {
       label: PropTypes.string,
       name: PropTypes.string,
       primary: PropTypes.bool,
+      defaultInternal: PropTypes.bool,
     })),
   };
 
+  renderVisibility = (internal, term) => {
+    if ((internal === false) || (!internal && term.defaultInternal === false)) {
+      return <span data-test-term-visibility={term.name}>{<FormattedMessage id="stripes-erm-components.term.internalFalse" />}</span>;
```
doytch: Let's break the two spans in this method into multiple lines. Also, it doesn't look like we need the braces around the `FormattedMessage`s

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -19,9 +19,18 @@ export default class LicenseTermsList extends React.Component {
       label: PropTypes.string,
       name: PropTypes.string,
       primary: PropTypes.bool,
+      defaultInternal: PropTypes.bool,
     })),
   };
 
+  renderVisibility = (internal, term) => {
+    if ((internal === false) || (!internal && term.defaultInternal === false)) {
```
doytch: Shortcutting with `!internal` is dangerous here since `internal` can be `false` as well as `undefined`. Additionally, the intent of this line could be difficult to understand for a new developer who doesn't have the full background of the various cases involved in calculating term visibility.

lib/LicenseTermsList/LicenseTermsList.js
```diff
@@ -19,9 +19,18 @@ export default class LicenseTermsList extends React.Component {
...
+      return <span data-test-term-visibility={term.name}>{<FormattedMessage id="stripes-erm-components.term.internalFalse" />}</span>;
+    }
+
+    return <span data-test-term-visibility={term.name}>{<FormattedMessage id="stripes-erm-components.term.internalTrue" />}</span>;
```
doytch: Finally, you can simplify this method by always returning the same span and FormattedMessage but just varying the `id` you pass into the FormattedMessage. Eg,
```
const internalId = ((internal === false) || (internal === undefined && term.defaultInternal === false)) ?
  'internalFalse' : 'internalTrue';

return (
  <span data-test-term-visibility={term.name}>
    <FormattedMessage id={`stripes-erm-components.term.${internalId}`} />
  </span>
);
```

lib/LicenseTermsList/tests/LicenseTermsList-test.js
```diff
@@ -102,10 +103,18 @@ describe('LicenseTermsList', () => {
     expect(licenseTermsList.licenseTerm(0).value).to.equal(get(license, ['customProperties', terms[0].name, '0', 'value']));
   });
 
-  it('renders the expected note under the first license term', () => {
+  it('renders the expected visibility under the first license term', () => {
+    expect(licenseTermsList.licenseTerm(0).internal).to.equal(get(license, ['customProperties', terms[0].name, '0', 'internal']));
```
doytch: I don't see the `internal` property defined in the interactor file. Am I missing something?

CalamityC:
> Nice work! Couple of comments regarding style, but the main thing is that I think the logic about showing/hiding doesn't match the JIRA. From my reading of ERM-355 and [this mockup](https://issues.folio.org/secure/attachment/20001/20001_License+preview+pane%2C+Terms+accordion.png), it seems like the `Visibility` should always be shown, and the `Public note` should only be shown if `Visibility` === `Public`.

No, from the JIRA issue: 
> The public note is always shown, even if blank (to indicate that no public note has been set)
And the Visibility is always displayed

doytch:
I believe the purpose of that is that if it's blank, don't just hide it, show it with a dash. But from the JIRA:

> When Visibility=Public and the "Public note" field is blank, display the field label and a "-".

Additionally, that interpretation actually matches the wireframe. Can you clarify this with Gill and Owen.

CalamityC:
I did already. I'm sure the behavior is correct. Empty public note is shown with a dash. In the wireframe the second term could indicate that the public note should be completely hidden if empty.

CalamityC:
I made the changes and hope it's good now. Thanks for your patience.

doytch:
> In the wireframe the second term could indicate that the public note should be completely hidden if empty.

I don't think so, because the third term is that case. The term's visibility is Public but no Public Note is set, so a dash is shown. OTOH, the second term's visibility is Internal so the notion of a Public Note doesn't make sense (the term is not visible to the public, so there's no need to set a note to _show_ the public) which is why it's not shown.

CalamityC:
Ok, you won. I feel brainwashed.

PR #101, opened 2019-08-12
ERM-309: Render Internal Contacts as Cards

doytch:
- Switch to the Card-based UX pattern for rendering contacts.
- Implementation is most similar to the Usage Providers implementation. I've been happiest with the separation of concerns of there, though by no means am I ecstatic about so please leave suggestions if you have any.
  - One of the things that doesn't feel great but I don't have better solutions is the `onMarkforDeletion ` business happening when we unlink a user. It doesn't feel clean...I dunno.
  - I feel rendering the `EditCard` in place is important to avoid mucking around with `key` but it's a hidden optimization and someone refactoring this might run into the issues that happen with orgs or linked licenses.
- Ditched `UserPicker` and wrote a new `UserField` component to replace it.
- Cleaned up the `helpers` by moving them all into separate files in a new `helpers` directory.
- Rewrote the `ICFA` tests by **mounting with a plugin**. 
  - This was novel to me. I didn't previously know how to test something that renders a `Pluggable` and thought it was impossible. Well it's not!

Misc
- Added `onPrependField` prop for components wrapped in a `withKiwtFieldArray`. We'll be using this for ERM-83 and ERM-391
- Bumped requirement on `eslint-config-stripes` to the newly-released `4.2.0`. Never again will we accidentally merge `describe.only` tests :)


lib/InternalContactsFieldArray/tests/InternalContactsFieldArray-test.js
```diff
@@ -33,9 +64,9 @@ describe('InternalContactsFieldArray', () => {
           name="InternalContactsFieldArrayTest"
           contactRoles={contactRoles}
           users={users}
-          userPickerComponent={TextField}
         />
-      </TestForm>
+      </TestForm>,
+      [findUserModule]
```
aditya-matukumalli: nice!

lib/InternalContactsFieldArray/UserField.js
```diff
@@ -0,0 +1,160 @@
...
+      onClick={this.props.onUserUnselected}
+    >
+      <FormattedMessage id="stripes-erm-components.contacts.unlinkUser" />
+    </Button>
+  )
+
+
+  renderUser = () => {
+    const {
+      email = '-',
```
aditya-matukumalli: Oo nice. Didn't know we could do this.

lib/InternalContactsFieldArray/UserField.js
```diff
@@ -0,0 +1,160 @@
...
+      firstName = '-',
+      lastName = '-',
+      phone = '-',
+    } = get(this.props.user, 'personal', {});
+
+    return (
+      <div data-test-user-card>
+        <Row>
+          <Col xs={6} md={3}>
+            <KeyValue label="Last name">
```
aditya-matukumalli: Assuming this is still WIP. Translations I mean.

doytch: Yup yup. Good catch though :D

lib/InternalContactsFieldArray/UserField.js
doytch: Yeah! That and the `:` renaming while destructuring really make me love destructuring overall.

PR #125, opened 2019-10-01
ERM-489: Migrate to react-final-form

doytch:
Begins a new major version, `2.0.0` because of the breaking changes to:
- DocumentsFieldArray
- InternalContactsFieldArray
- EditOrganizationCard

All of those components must now be used inside a `react-final-form` form rather than a `redux-form` form.


doytch:
Marking this as `do-not-merge` for now. We want to merge agreement periods and new permissions before this.

PR #130, opened 2019-10-15
Tweak erm components to accept tooltip prop

aditya-matukumalli:


lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -162,6 +164,42 @@ export default class EditOrganizationCard extends React.Component {
     );
   }
 
+  renderDeleteButton = () => {
```
doytch: Can we use an EditCard here instead of duplicating this logic?

lib/OrganizationsFieldArray/OrganizationsFieldArray.js
```diff
@@ -11,74 +11,75 @@ import OrganizationPicker from './OrganizationPicker';
...
+  renderOrganizations = (items) => {
+    const orgs = items.map((organization, index) => {
       return (
-        <div>
-          {orgs}
-        </div>
+        <EditOrganizationCard
+          index={index}
+          key={index}
+          deleteButtonTooltipText={<FormattedMessage id="stripes-erm-components.organizations.removeOrganization" />}
```
doytch: This FormattedMessage would preferably just be defined in the card component itself since it doesn't look like we'll need to customise it from outside of that component.

lib/EditCard/EditCard.js
```diff
@@ -16,16 +15,37 @@ export default class EditCard extends React.Component {
...
+  renderDeleteButton = () => {
+    const { deleteButtonTooltipText } = this.props;
+    return (
+      deleteButtonTooltipText ?
+        (
+          <Tooltip
+            text={deleteButtonTooltipText}
+          >
+            {({ ref, ariaIds }) => (
+              <IconButton
```
doytch: Can you clean this section up a bit? In one case we're rendering an IconButton and in the other a `<Button><Icon /></Button>`. Go ahead and just use an IconButton all the time since it looks like that way we avoid the blue colour that Gill requested we remove.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change history for stripes-erm-components
 
+## 2.1.0 2019-10-16
```
doytch:
```suggestion
## 2.1.0 IN PROGRESS
```

PR #147, opened 2019-11-22
Disable show credentials button when the user does not have permissio…

aditya-matukumalli:
…n, refs ERM-493

- Disabled the _show credentials_ button when the user doesn't have the necessary permission. 
- Added a _ToolTip_ that displays an appropriate message when the button is disabled. 
- Used _IfPermission_ to conditionally render

Note: Had to render the Button with the ToolTip inside a **div** and `ref` ed on it, as I believe that any pointer-events on a disabled button are set to `none` FOLIO-wide.


lib/ViewOrganizationCard/ViewOrganizationCard.js
```diff
@@ -98,16 +99,36 @@ export default class ViewOrganizationCard extends React.Component {
...
+                  >
+                    {({ ref, ariaIds }) => (
+                      <div ref={ref} aria-labelledby={ariaIds.text}>
+                        <Button
+                          data-test-show-credentials
+                          buttonStyle="primary"
+                          disabled
+                          marginBottom0
+                        >
+                          {<FormattedMessage id="stripes-erm-components.interface.showCredentials" />}
```
doytch:
```suggestion
                          <FormattedMessage id="stripes-erm-components.interface.showCredentials" />
```

PR #153, opened 2019-12-02
ERM-625: Maximum file size message needs updating

CalamityC:
update en.json, en_US.json


doytch:
Apparently there's uncertainty about whether the limit should be 100 MB or 200 MB (since that's what it is on the backend). Will leave this unmerged for a bit pending response from Owen and Gill.

PR #164, opened 2019-12-10
ERM-603 and ERM-604: PR to change the way organization cards are rendered

EthanFreestone:
This is a little bit of work to change the way that the name of an organization is rendered on an organization card.

I've had some problems getting my agreements form to work, so I've been unable to test whether all tests pass on ui-agreements, but there seem to be no issues in stripes-erm-components or ui-licenses using this, at least for me locally.

### DIFF
--- a/lib/ViewOrganizationCard/ViewOrganizationCard.js
+++ b/lib/ViewOrganizationCard/ViewOrganizationCard.js
@@ -6,8 +6,10 @@ import { get, isEmpty } from 'lodash';
 import {
   Button,
   Card,
+  Col,
   Icon,
   MultiColumnList,
+  Row,
   Tooltip,
 } from '@folio/stripes/components';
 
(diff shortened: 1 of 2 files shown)


lib/ViewOrganizationCard/ViewOrganizationCard.js
```diff
@@ -139,16 +141,16 @@ export default class ViewOrganizationCard extends React.Component {
...
     const { id } = this.props;
     return (
-      <span>
-        <AppIcon
-          app="organizations"
-          size="small"
-        >
-          <strong id={id} data-test-org-name>
-            {orgName}
+      <Col>
```
doytch: Cols need to be children of Rows, not vice versa.

lib/ViewOrganizationCard/ViewOrganizationCard.js
```diff
@@ -139,16 +141,16 @@ export default class ViewOrganizationCard extends React.Component {
...
+        <Row>
+          <strong>
+            Name
```
doytch: This needs to be a translated string.

lib/ViewOrganizationCard/ViewOrganizationCard.js
EthanFreestone: Whoops, I've switched them now

lib/ViewOrganizationCard/ViewOrganizationCard.js
EthanFreestone: Yeah, I missed that one, cheers

PR #168, opened 2019-12-12
Give focus to first field in a Document field, refs ERM-575

aditya-matukumalli:
Factored out `DocumentsFieldArray`s `Field`s into a separate component  to facilitate focus on mount on the first field


lib/DocumentsFieldArray/DocumentsFieldArray.js
```diff
@@ -117,79 +57,14 @@ class DocumentsFieldArray extends React.Component {
...
-                this.validateURLIsValid,
-              )}
-            />
-          </Col>
-        </Row>
+        <Field
+          component={DocumentField}
+          documentCategories={documentCategories}
+          index={i}
+          name={`${name}`}
```
doytch: OK wait a sec here, shouldn't this be something like the following?
```suggestion
          name={`${name}[${i}]`}
```

That'll allow you to avoid [this weird indexing](https://github.com/folio-org/stripes-erm-components/pull/168/files#diff-556c68713cd130583f07663def58a56fR40).

aditya-matukumalli: Yeah you are right. Although, this would need a lot of changes in the tests I believe, such as changing all the selectors names https://github.com/folio-org/ui-agreements/blob/master/test/ui-testing/docs.js#L52. So its either this vs that?

PR #180, opened 2020-01-07
PR for ERM-629

EthanFreestone:
This deals with the stripes-erm-components half of ERM-629, including changes to allow indexing of the "remove" button tooltips, and adding a tooltip with specific information for "replace" buttons


lib/InternalContactsFieldArray/UserField.js
```diff
@@ -49,18 +56,40 @@ export default class UserField extends React.Component {
       selectUser={this.props.onUserSelected}
       renderTrigger={(props) => {
         this.triggerButton = props.buttonRef;
+
+        const userString = this.props.user.personal ? `${this.props.user.personal.lastName}${this.props.user.personal.firstName ? ', ' : ' '}${this.props.user.personal.firstName ? this.props.user.personal.firstName : ''} ${this.props.user.personal.middleName ? this.props.user.personal.middleName : ''}` : `user ${this.props.index + 1}`;
```
doytch: We've got a bunch of these statements littered throughout the library now. Ideally we'd just define this in one place and import it everywhere we need to get a user's name as a string. 

I'll make a pass through after this is merged and clean it up.

lib/OrganizationsFieldArray/EditOrganizationCard.js
```diff
@@ -139,22 +145,48 @@ export default class EditOrganizationCard extends React.Component {
               renderTrigger={(props) => {
                 this.triggerButton = props.buttonRef;
 
+                let orgName = '';
```
doytch: It'd be cleaner to just `const org = get(this.props.organization, 'org.name');`

lib/DocumentsFieldArray/DocumentsFieldArray.js
```diff
@@ -15,7 +15,10 @@ import DocumentField from './DocumentField';
 class DocumentsFieldArray extends React.Component {
   static propTypes = {
     addDocBtnLabel: PropTypes.node,
+    // TODO - make the following line work
+    // deleteButtonTooltipText: deprecated(PropTypes.node, `Use the "deleteBtnTooltipMsgId" to pass in a translation string instead.`),
```
doytch: If we want to deprecate it we _need_ to warn the user. So this either needs to work, or you need to do the console.warn method in the constructor.

EthanFreestone: Ah, thank you, I'd legitimately forgotten that I hadn't finished that. Every time I added the `deprecated` bit I ended up with the front end crashing, is there a piece of the process I'm missing? I did add it to the `package.json`

doytch: You did import it in this file, right? Something like `import { deprecated } from 'prop-types-extra';`?

lib/OrganizationsFieldArray/EditOrganizationCard.js
EthanFreestone: This has now been changed, along with other similar instances

lib/DocumentsFieldArray/DocumentsFieldArray.js
EthanFreestone: Ah, I had forgotten to do that, I'll blame Christmas cobwebs. Cheers Mark.

PR #185, opened 2020-01-29
add new requiredObjectValidator refs ERM-569

CalamityC:


lib/validators.js
```diff
@@ -5,6 +5,15 @@ const required = value => (
   (value || value === false || value === 0) ? undefined : <FormattedMessage id="stripes-core.label.missingRequiredField" />
 );
 
+const requiredObject = (value = {}) => {
+  // withKiwtFieldArray sets the _delete property on new objects by default
+  delete value._delete;
```
doytch: Ooooops, I was wrong. I didn't catch when I gave you this option that it's actually mutating the object that's being passed into the function. So after this function runs, the `_delete` property will continue to be stripped in react-final-form. That's a problem! 🤦‍♂ 

We should definitely go with the alternative approach of plucking out the _delete and using the ...rest prop as the object we'll check. Eg, `const value = { _delete, ...formValue };`

PR #186, opened 2020-01-30
Move FormCustomProperties into its own ERM component

aditya-matukumalli:
Added new components:
- FormCustomProperties (includes CustomPropertiesListField,  (lists custom properties in edit page))
- CustomPropertiesConfigListFieldArray (settings)
- CustomPropertiesList (lists custom properties in view page)

Conditionally render pickLists dropdown only when used in licenses.

Renamed LicenseTermsList to CustomPropertiesList


doytch:
Like the other PR, can we not call this "terms" since that's only a specific use case of these components? Also, let's get tests for the rest of the components and to get the coverage up.

lib/CustomPropertiesConfig/CustomPropertiesConfigForm.js
```diff
@@ -25,21 +28,31 @@ class TermsConfigForm extends React.Component {
...
+    return (
+      this.callout.sendCallout({
+        type: 'error',
+        message: <SafeHTMLMessage id={translationKey} />,
+        timeout: 0,
+      })
+    );
+  }
+
+  getSettingsRoute = () => {
```
doytch: This feels pretty flaky. Can we just pass in `terms` or `customProperties` as a prop and use that instead?

lib/CustomPropertiesConfig/CustomPropertyFieldEdit.js
```diff
@@ -37,32 +38,35 @@ export default class TermFieldEdit extends React.Component {
       onCancel,
       onSave,
       pickLists,
+      settingsRoute
     } = this.props;
 
+    const translationKey = (settingsRoute === 'terms') ? 'Term' : 'SupplementaryProperty';
```
doytch: Following on from the other component's comment about using the URL, let's rename this to something more generic. If we end up passing the translation key fragment directly to `CustomPropertiesConfigForm`, we can probably just pass that down here. And maybe clean up the translation keys so we don't sometimes use `terms` and sometimes use `Term` (L55 vs L86)? Just use the opportunity we have now to clean things up.

lib/CustomPropertiesConfig/CustomPropertyFieldView.js
```diff
@@ -37,25 +38,31 @@ export default class TermFieldView extends React.Component {
       input: { value = {} },
       onDelete,
       onEdit,
+      settingsRoute,
     } = this.props;
 
+    const translationKey = (settingsRoute === 'terms') ? 'Term' : 'SupplementaryProperty';
```
doytch: Same comment as on CPFieldEdit.

lib/CustomPropertiesList/CustomPropertiesList.js
```diff
@@ -23,21 +23,21 @@ export default class CustomPropertiesList extends React.Component {
...
+      <span data-test-customproperty-visibility={customProperty.name}>
+        <FormattedMessage id={`stripes-erm-components.customProperty.${internalId}`} />
       </span>
     );
   }
 
-  renderTerm = (term, index) => {
-    let value = get(this.props.resource, ['customProperties', term.name, '0', 'value']);
+  renderCustomProperty = (customProperty, index) => {
+    let value = get(this.props.resource, ['customProperties', customProperty.name, '0', 'value']);
```
doytch: Let's take opportunities like this Issue/PR to reduce our usage of `get`

translations/stripes-erm-components/en.json
```diff
@@ -102,49 +102,69 @@
   "errors.uploadError.1": "<strong>Error:</strong> The file was not uploaded because the connection was dropped.",
   "errors.uploadError.2": "Please check the filesize is <strong>{number}MB or less</strong> and try again.",
   "errors.uploadError.3": "If the problem persists contact your system administrator.",
-  "errors.termNameHasNonAlpha": "The term name can only contain Latin alphabetical characters (a-z) and non-leading numeric characters.",
+  "errors.hasNonAlphaTermName": "The term name can only contain Latin alphabetical characters (a-z) and non-leading numeric characters.",
```
doytch: The following isn't a big deal, I mention it because of the previous comments about terms/Terms:

Having separate translations for a message like this isn't really important. We could remove the "term" from this message and it's still completely obvious what we're talking about because it's next to the `Name` field. So if we only have an `errors.hasNonAlphaName` translation, we don't need to check which translation to render based on whether we're rendering terms/custom properties. So it may actually let you avoid having to pass props or do the `terms -> Terms` conversion, etc, basically cleaning things up.

lib/CustomPropertiesConfig/CustomPropertiesConfigForm.js
aditya-matukumalli: yeah, I am working on it locally and have made that change as I now moved the `CustomPropertiesConfigForm` to the respective apps (from where I pass in the `terms` or `supplementaryProperties`) and just have the `CustomPropertiesConfigListFieldArray` as a shared component

translations/stripes-erm-components/en.json
aditya-matukumalli: Will cleanup as a part of the final skim through the PR. Thanks

PR #211, opened 2020-03-13
cleanup validation messages, refs erm-768

aditya-matukumalli:


lib/FormCustomProperties/CustomPropertiesListField.js
```diff
@@ -141,11 +141,19 @@ export default class CustomPropertiesListField extends React.Component {
       const { note, publicNote, value } = val ? val[0] : {};
 
       if (!customProperty.primary && customProperty.value && !value) {
-        errors[customProperty.value] = <FormattedMessage id="stripes-core.label.missingRequiredField" />;
+        if (customProperty.type === CUSTOM_PROPERTY_TYPE_NUMBER || customProperty.type === CUSTOM_PROPERTY_TYPE_DECIMAL) {
```
doytch: While we're doing this, can you change this and any other references to these constants to pull from `customPropertyTypes.js` instead?

lib/FormCustomProperties/CustomPropertiesListField.js
```diff
@@ -141,11 +141,19 @@ export default class CustomPropertiesListField extends React.Component {
       const { note, publicNote, value } = val ? val[0] : {};
 
       if (!customProperty.primary && customProperty.value && !value) {
```
doytch: Will this be true and throw an error when the valid number/decimal `value` is `0`?

aditya-matukumalli: See comment here -> https://github.com/folio-org/ui-licenses/pull/320/files#r393036097

lib/FormCustomProperties/CustomPropertiesListField.js
```diff
@@ -216,18 +198,51 @@ export default class CustomPropertiesListField extends React.Component {
...
-        error={!isEmpty(errors) ? errors[customProperty.value] : undefined}
-        id={`edit-customproperty-${i}-value`}
-        label={<FormattedMessage id="stripes-erm-components.prop.customPropertyValue" />}
-        onChange={handleChange}
-        value={controlledFieldValue}
-        {...fieldProps}
-        required={!customProperty.primary}
-      />
-    );
+    return (customProperty.type === customPropertyTypes.NUMBER || customProperty.type === customPropertyTypes.DECIMAL) ?
```
doytch: This can be done a lot cleaner a few different ways that we already make use of. Eg, grouping shared props and spreading them like with `fieldProps`.

Given the size that this function is getting to, it might also be worth breaking it out into its own component. Then we wouldn't need to do an IntlConsumer wrap on one part of a ternary expression either (though I really would like to avoid this ternary in general).

lib/FormCustomProperties/CustomPropertiesListField.js
```diff
@@ -216,18 +198,51 @@ export default class CustomPropertiesListField extends React.Component {
...
+      (
+        <IntlConsumer>
+          { intl => (
+            <Field
+              badInput={intl.formatMessage({ id: 'stripes-erm-components.errors.customPropertyInvalidNumber' })}
+              component={TextField}
+              data-test-customproperty-value
+              inputRef={this.inputRef}
```
doytch: This `inputRef` isn't doing anything anymore is it?

lib/FormCustomProperties/tests/FormCustomPropertiesInteractor.js
```diff
@@ -36,17 +37,21 @@ import {
...
-  value = text('[data-test-customproperty-value]');
+  value = value('[data-test-customproperty-value]');
   visibility = value('[data-test-customproperty-visibility]');
 
   nameField = scoped('[data-test-customproperty-name]', NameFieldInteractor);
   valueField = scoped('[data-test-customproperty-value]', ValueFieldInteractor);
   visibilityField = scoped('[data-test-customproperty-visibility]', visibilityFieldInteractor);
 
   fillNote = fillable('[data-test-customproperty-note]');
+  blurNote = blurrable('[data-test-customproperty-note]');
```
doytch: The TextField interactor has a `fillAndBlur` interactor you might be able to use for the `fill*` interactors in place of `fillable`. That would let you avoid the `.fill().blur()` chaining you do in the tests.

lib/FormCustomProperties/CustomPropertiesListField.js
aditya-matukumalli: No, it isnt.

lib/FormCustomProperties/CustomPropertiesListField.js
aditya-matukumalli: Yeah, was going to clean it up, but yeah will move into a different component so that `intl` can be passed down

lib/FormCustomProperties/tests/FormCustomPropertiesInteractor.js
aditya-matukumalli: I tried it that way...seems it cant find the `input` that is being targetted because the `data-test-*` attributes are on the input? I might have to move the `data-test-*` atttribute onto a div/Fragment that wraps the new component I create above to target the value Field. Not sure if I can grab the Select and Note via the `data-test`, might have to tweak it.....will try

PR #220, opened 2020-04-06
Added DuplicateModal shared component, refs ERM-814

aditya-matukumalli:
- Moved the clone functionality across agreements and license into a shared component
- Tests

### DIFF
--- a/lib/DuplicateModal/DuplicateModal.js
+++ b/lib/DuplicateModal/DuplicateModal.js
@@ -1,0 +1,176 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { keyBy, mapValues, pickBy } from 'lodash';
+import { FormattedMessage } from 'react-intl';
+import {
+  Button,
+  Checkbox,
+  Icon,
+  Layout,
+  Modal,
+  ModalFooter,
+} from '@folio/stripes/components';
+
+import css from './DuplicateModal.css';
+import errorTypes from '../errorTypes';
+
+export default class DuplicateModal extends React.Component {
+  static propTypes = {
+    onClose: PropTypes.func,
+    onClone: PropTypes.func,
+    translationKey: PropTypes.string,
+    cloneableProperties: PropTypes.arrayOf(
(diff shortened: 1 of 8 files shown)


lib/DuplicateModal/DuplicateModal.js
```diff
@@ -0,0 +1,176 @@
...
+} from '@folio/stripes/components';
+
+import css from './DuplicateModal.css';
+import errorTypes from '../errorTypes';
+
+export default class DuplicateModal extends React.Component {
+  static propTypes = {
+    onClose: PropTypes.func,
+    onClone: PropTypes.func,
+    translationKey: PropTypes.string,
```
doytch, line 21: The way this is plopped directly in the translation string means that [passing it directly here](https://github.com/folio-org/ui-agreements/pull/481/files#diff-39afc5530cdd563ebdf087b08d317200R22) prevents it from being translated. Eg, in Japanese, it might say `複製 agreement`. 😬

aditya-matukumalli: Ah right... will translate it within the component before being used then

`const translationKey = `stripes-erm-components.${props.translationKey}`

doytch: I'm really hesitant about that solution but don't have a better one off the top of my head. I'd suggest bringing this to a wider audience (eg, in #stripes).

Many languages have forms where the way you use a noun changes (declension) based on the way its used in a sentence. Eg, in Czech, "look at the castle" and "they're in the castle" have different ways of spelling "castle" ("hrad"  vs "hradu"). 

So unless the `translationKey` is only used in one place and guaranteed to be in one usage/form, it's pretty unsafe.

aditya-matukumalli:
This PR is progressable now that https://github.com/folio-org/mod-licenses/pull/130 has been merged. The corresponding PRs in [agreements](https://github.com/folio-org/ui-agreements/pull/481) and [licenses](https://github.com/folio-org/ui-licenses/pull/333) needs to be merged after this PR is handled

PR #221, opened 2020-04-09
ERM-827 Add support for "Alternative name" for agreements

CalamityC:
* add AlternateNamesFieldArray, also refs ERM-828


CalamityC:
> 
> Super nitpick, but if we're calling this "alternative" names in the UI, can we call it that in code as well instead of "alternate" names?

I changed all occurrences in the frontend. But I leave the property's name in the backend as it is.

> Also, some unit tests to confirm that everything's wired up properly and is submitting correctly would be good. You can take inspiration from the withKiwtFieldArray tests that use chai spies to check what they're being passed.

Ok

doytch:
> > Super nitpick, but if we're calling this "alternative" names in the UI, can we call it that in code as well instead of "alternate" names?
> 
> I changed all occurrences in the frontend. But I leave the property's name in the backend as it is.

Thanks. Is it intentional that some of the `id`s and what not are still using `alternate`? I'm just hesitant because I can totally see myself going into the code later on, introducing a bug, and banging my head against the keyboard because I used "alternative" instead of "alternate" or vice-versa. 😆

CalamityC:
> Thanks. Is it intentional that some of the `id`s and what not are still using `alternate`? I'm just hesitant because I can totally see myself going into the code later on, introducing a bug, and banging my head against the keyboard because I used "alternative" instead of "alternate" or vice-versa. 😆
No, ids containing _alternate_ are not intentional. But that's still work in progress, so I'll change that.
anyway thanks for the hint

CalamityC:
I basically copied the `withKiwtFieldArrayTest` and tried to adapt it.

lib/AlternativeNamesFieldArray/AlternativeNamesFieldArray.js
```diff
@@ -0,0 +1,82 @@
...
+          key={index}
+          data-test-render-alternative-names
+        >
+          <Row>
+            <Col xs={11}>
+              <Field
+                autoFocus
+                component={TextField}
+                data-test-alternative-names-field
+                index={index}
```
doytch, line 42: Does this do something?

lib/AlternativeNamesFieldArray/AlternativeNamesFieldArray.js
```diff
@@ -0,0 +1,82 @@
...
+      const { items, name, onDeleteField } = this.props;
+      return items.map((item, index) => (
+        <div
+          key={index}
+          data-test-render-alternative-names
+        >
+          <Row>
+            <Col xs={11}>
+              <Field
+                autoFocus
```
doytch, line 39: Are you sure about this? What happens when you open up an agreement for editing that already has some alternative names? We _shouldn't_ be autofocusing this field.

lib/AlternativeNamesFieldArray/tests/AlternativeNamesFieldArray-test.js
```diff
@@ -0,0 +1,204 @@
...
+import { Form } from 'react-final-form';
+import arrayMutators from 'final-form-arrays';
+import { FieldArray } from 'react-final-form-arrays';
+
+import { Button } from '@folio/stripes/components';
+
+import { dummyMount, setupApplication } from '../../../tests/helpers';
+import AlternativeNamesFieldArray from '../AlternativeNamesFieldArray';
+
+import TasksFormInteractor from './interactor';
```
doytch, line 16: What's all this about tasks in this file? Please be careful with copy-and-paste. 🙃

CalamityC: Hm. Isn't it just a name? Again, laziness. Do you want me to change all the `task` names into `name` or `alternativeName` for Instance?

doytch: It's incredibly frustrating to have to parse through new code when parts of it are irrelevant, since I have to hold in my head the parts that are relevant and discard the parts that aren't. I didn't even bother to look at these tests because it's just such an annoying and time-consuming thing to do.

And whenever someone works on those tests, they'll be repeating what you'd make me do right now if I was going to review it.

CalamityC: you leave me perplexed.
i don't think anything is irrelevant. but i'm not good enough at bigtest, so i have to copy code and try to adapt. In my opinion I just didn't change the names, I can still do that

lib/AlternativeNamesFieldArray/AlternativeNamesFieldArray.js
```diff
@@ -0,0 +1,80 @@
...
+    renderNames = () => {
+      const { items, name, onDeleteField } = this.props;
+      return items.map((item, index) => (
+        <div
+          key={index}
+          data-test-render-alternative-names
+        >
+          <Row>
+            <Col xs={11}>
+              <Field
```
doytch, line 38: Sorry for not being super clear on the autofocus part. It's important to not _always_ have autofocus on the Field, but we do need to have it focus when a new alternative name is added. 

Eg, note how when you add a new document on a section with DocumentsFieldArray, the Name field is automatically focused instead of focus remaining with the `Add supplementary document` (or similar) button?

We do it by separating out the `DocumentField` parts into their own component, and when that component mounts, [we check whether we should autofocus a particular element](https://github.com/folio-org/stripes-erm-components/blob/5034b0bc4379698ed054feadc0e986ae6c98b280/lib/DocumentsFieldArray/DocumentField.js#L38). In that `if` case, we're checking whether the document object we're editing has an `id`. If it does, it was generated by the backend when that document was initially saved. If it doesn't have an `id` and the component mounted, we can then conclude that the document was added right now by the user pressing an **Add document** button.

So similarly, you'll need to break this component out into an `AlternativeNamesField` and `AlternativeNamesFieldArray` component and use similar `this.inputRef` and `componentDidMount` logic to autofocus the textfield, but _only_ if it's being mounted because the user _just_ added the alternative name, and not if we're loading the Edit Agreement page.

PR #236, opened 2020-05-12
ERM-876 Add ability to search/filter agreements based on custom properties

CalamityC:
* transfer TermsFilters from ui-licenses to CustomPropertyFilters
* add test for CustomPropertyFilters


lib/CustomPropertyFilters/CustomPropertyFilters.js
```diff
@@ -0,0 +1,89 @@
...
+        .join('||'));
+
+    filterHandlers.state({ ...activeFilters, custProps: filterStrings });
+
+    return Promise.resolve();
+  }
+
+  render() {
+    const { activeFilters, data, filterHandlers, custPropName } = this.props;
+    const custProps = data?.terms ?? data?.supplementaryProperties;
```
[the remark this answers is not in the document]
CalamityC: Yes I thought of that also. And then don't pass the whole `data` at all?

lib/DocumentsFieldArray/DocumentField.js
```diff
@@ -86,7 +86,7 @@ export default class DocumentField extends React.Component {
             ]}
             id={`${id}-category`}
             label={<FormattedMessage id="stripes-erm-components.doc.category" />}
-            name={`${name}.atType`}
+            name={`${name}.atType.value`}
```
doytch, line 89: I'm not certain about this. We normally don't use this approach because [we process the record before providing it as `initialValues`.](https://github.com/folio-org/ui-licenses/blob/9e73260a51b06c2d154ec84bb9ce4ac040dddd59/src/routes/EditLicenseRoute.js#L130). 

I'm assuming this fix will only work in Agreements but will break Licenses. If so, can you instead fix this in Agreements by transforming the docs objects (that have doc types) instead?

lib/CustomPropertyFilters/CustomPropertyFilters.js
```diff
@@ -66,9 +65,9 @@ export default class CustomPropertyFilters extends React.Component {
         closedByDefault
         displayClearButton={numberOfFilters > 0}
         header={FilterAccordionHeader}
-        id="clickable-custprops-filter"
+        id="clickable-customProperties-filter"
```
doytch, line 68: Sorry, but unless we have a really strong reason for it, prefer kebab-casing `id`s.
```suggestion
        id="clickable-custom-properties-filter"
```

lib/CustomPropertyFilters/CustomPropertyFiltersForm.js
```diff
@@ -0,0 +1,196 @@
...
+        change,
+      },
+      values,
+    } = this.props;
+    const { editingFilters } = this.state;
+
+    return (
+      <>
+        <Button
+          data-test-open-customProperties-filters
```
doytch, line 88: Looks like the tests are breaking because this data-test attribute is different from what's expected. Can you do a pass through this PR and make sure we're using the same format across all these data-test attrs and ids? Ie, `custom-properties` vs `custprops` vs `customProperties`?

CalamityC: Yes. But I decided on `custprop` instead of `custom-properties` because almost all of the ids and attributes used this already

PR #250, opened 2020-06-09
Revert "Helper function to unmount a component not mounted via setupApplication"

aditya-matukumalli:
Reverts folio-org/stripes-erm-components#249 to prevent 

Module not found: Error: Can't resolve '@bigtest/react' in '/home/jenkins/workspace/ReShare Automation/reshare-dev-frontends/rs_ui/node_modules/@folio/stripes-core/test/bigtest/helpers'"


skomorokh:
It's not just a ReShare problem, we just noticed it first. By putting that in index you have this depending on devDependencies which I'm pretty sure will create a problem for anything that brings it in via npm, including FOLIO platforms.

PR #254, opened 2020-06-19
Added shared utility helpers

aditya-matukumalli:
- Added `EResourceType`, `getResourceIdentifier`, `getSiblingIdentifier` and `isPackage` utility functions used across agreements and plugin-find-eresource.
- Needs to be added in before merging https://github.com/folio-org/ui-plugin-find-eresource/pull/1


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change history for stripes-erm-components
 
+## 3.1.0 IN PROGRESS
```
doytch, line 3: I like to bump the version in package.json at this point. Allows you to depend on `^3.1.0` in whatever apps you use these features, and ensures you won't accidentally release a package that has this unreleased dependency.

index.js
```diff
@@ -30,8 +30,11 @@ export { default as withKiwtFieldArray } from './lib/withKiwtFieldArray';
 
 // Functions, utilities, and misc.
 export { default as customPropertyTypes } from './lib/customPropertyTypes';
+export { default as EResourceType } from './lib/EResourceType';
```
doytch, line 33: Components go in the first block per file style.

doytch:
It'd be a good idea to write tests for these utilities. These things that pull in data from various sources are tough to test outside of unit tests. Additionally, we're already planning on refactoring this stuff in the future so tests allow us to confidently refactor without being afraid that we're missing something.

index.js
aditya-matukumalli: > It'd be a good idea to write tests for these utilities. These things that pull in data from various sources are tough to test outside of unit tests. Additionally, we're already planning on refactoring this stuff in the future so tests allow us to confidently refactor without being afraid that we're missing something.

Sounds good

aditya-matukumalli: Updated the PR with utility helpers tests.  Just FYI: `getResourceIdentifier` and ``getSiblingIdentifier` are currently being called in only one [context](https://github.com/folio-org/ui-agreements/blob/master/src/components/views/EResources.js#L231), whereas `EresourceType` is being called from 6 different contexts. All of them should be covered by these tests

PR #258, opened 2020-06-29
add note to organizations refs ERM-966, ERM-967

CalamityC:
* add org note to translations
* add org note to edit card
* add note to ViewOrganizationCard


CalamityC:
I added tests for the note field to existing test scripts. 
The backend is now merged.

PR #269, opened 2020-07-29
ERM-943 Separate ERM 'publication type' from 'type'

CalamityC:
* add publicationType property to EResourceType

### DIFF
--- a/lib/EResourceType/tests/interactor.js
+++ b/lib/EResourceType/tests/interactor.js
@@ -4,5 +4,5 @@ import {
 } from '@bigtest/interactor';
 
 export default @interactor class EresourceTypeInteractor {
-  publicationType = text('[data-test-eresource-type]');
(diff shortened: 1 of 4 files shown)


lib/EResourceType/tests/interactor.js
```diff
@@ -4,5 +4,5 @@ import {
 } from '@bigtest/interactor';
 
 export default @interactor class EresourceTypeInteractor {
-  type = text('[data-test-eresource-type]');
+  publicationType = text('[data-test-eresource-type]');
```
aditya-matukumalli, line 7: could we change the `data-test-*` attributes name from type to publicationType just to stay consistent. You would have to reflect this change at the app level as well

aditya-matukumalli, line 7: Could we change `data-test-eresource-type` to `data-test-eresource-publication-type`.

PR #285, opened 2020-09-22
ERM-1068 Agreements: Improve the display of 'interface' details for linked Organizations

CalamityC:
* use min and max width for MCL in ViewOrganizationCard


aditya-matukumalli (changes requested):
image
From the screenshot the Username and Password column headers are overlapping. So, the widths for those columns have to be adjusted, probably just add a `min` on each of those columns

CalamityC:
> image
> From the screenshot the Username and Password column headers are overlapping. 

That's weird, I didn't even touch that values. It's not overlapping in my environment. I changed it, please have a look if it's enough.

aditya-matukumalli (approved):

PR #289, opened 2020-10-05
ERM-1121 add alignment option to embargo component

CalamityC:
ERM-1121 add alignment option to embargo component


lib/Embargo/Embargo.js
```diff
@@ -5,6 +5,7 @@ import { Layout } from '@folio/stripes/components';
 
 export default class Embargo extends React.Component {
     static propTypes = {
+      alignment: PropTypes.string,
```
aditya-matukumalli, line 8: Following the comment [here](https://github.com/folio-org/ui-agreements/pull/629#pullrequestreview-503477149) you could change this proptype to be 
```suggestion
      alignment: PropTypes.oneOf(['left', 'right', 'center'])
```

lib/Embargo/Embargo.js
```diff
@@ -17,13 +18,13 @@ export default class Embargo extends React.Component {
       })
     }
```
aditya-matukumalli, line 20: Add a `defaultProp` here 

```suggestion
      static defaultProps = {
           alignment: 'left'
      };
```

lib/Embargo/Embargo.js
```diff
@@ -17,13 +18,13 @@ export default class Embargo extends React.Component {
...
 
-    renderEmbargoEnd(embargo) {
+    renderEmbargoEnd(embargo, className) {
       if (embargo.movingWallEnd?.length) {
         const unitId = `stripes-erm-components.${embargo.movingWallEnd.unit}`;
         const number = embargo.movingWallEnd.length;
         return (
           <Layout
-            className="flex centerContent"
+            className={className}
```
aditya-matukumalli, line 27: you could then change this to 
```suggestion
            className={css[this.props.alignment]}
```

A new css file `Embargo.css` needs to be added for this with the following classnames

```
.left {
    display: flex; 
    align-items: center
}

.right {
   display: flex; 
    justify-content: flex-end;
}

.center {
    display: flex;
    justify-content: center
}
```

Please double check the key-values from https://github.com/folio-org/stripes-components/blob/master/lib/Layout/stories/classNamePropertiesMapping.js

lib/Embargo/Embargo.js
```diff
@@ -36,13 +37,13 @@ export default class Embargo extends React.Component {
...
 
-    renderEmbargoStart(embargo) {
+    renderEmbargoStart(embargo, className) {
       if (embargo.movingWallStart?.length) {
         const unitId = `stripes-erm-components.${embargo.movingWallStart.unit}`;
         const number = embargo.movingWallStart.length;
         return (
           <Layout
-            className="flex centerContent"
+            className={className}
```
aditya-matukumalli, line 46: see above

lib/Embargo/Embargo.js
```diff
@@ -56,13 +57,15 @@ export default class Embargo extends React.Component {
     }
 
     render() {
-      const { embargo } = this.props;
+      const { alignment, embargo } = this.props;
```
aditya-matukumalli, line 60: This whole render block can be cleaned up with the above changes

aditya-matukumalli (changes requested):

lib/Embargo/Embargo.js
CalamityC: If we solve it like this, why can't the default alignment be center? Because it's the default now, we don't have to pass something to have the Embargo line centered. Couldn't we just leave it like this?

aditya-matukumalli: That make sense

aditya-matukumalli (approved):

PR #342, opened 2021-01-25
ERM-1532: Search by Agreement start / end date

EthanFreestone:
Expand DateFilter component to make more flexible. Accepts two new props:
- hideNoDateSetCheckbox - controls appearance of "include {resourceName} with start/enddate unset
- resourceName - string to be passed for labelling purposes


lib/DateFilter/DateFilter.js
```diff
@@ -52,18 +52,22 @@ const DateFilter = ({ activeFilters, filterHandlers, name }) => {
       separator={false}
     >
       <DateFilterForm
+        allowNull={allowNull}
```
aditya-matukumalli, line 55: https://github.com/folio-org/ui-agreements/pull/739#discussion_r563998922

aditya-matukumalli (changes requested):

lib/DateFilter/DateFilter.js
EthanFreestone: Change made, to `showNoDateCheckbox` (switched from showNoEndDateCheckbox since this component is covers both start and end dates)

PR #347, opened 2021-02-01
add handleSaveKeyCommand to keyboardShortcuts, refs ERM-1239, ERM-124…

CalamityC:
…0, ERM-1241


aditya-matukumalli:
Now that all the imports we use here are available from [stripes-components](https://github.com/folio-org/stripes-components/blob/master/index.js#L92-L95), lets cleanup the exports from stripes-erm-components. Note: `ui-agreements` also needs to import from stcom before cleaning up from erm-components

CalamityC:
> All the exprts for shortcut keys are available from [stripes-components](https://github.com/folio-org/stripes-components/blob/master/index.js#L92-L95), so lets cleanup the exports from stripes-erm-components. Note: `ui-agreements` also needs to import from stcom before cleaning up from erm-components

As we talked yesterday, could we do the cleanup work in a separate issue / branch? I need the handleSaveKeyCommand in stripes-erm-components to proceed with the keyboard shortcut issues in  ui-licenses, ui-local-kb-admin and ui-erm-comparisons. And we also need it there if we want to do the cleanup work in ui-agreements.

aditya-matukumalli (approved):

PR #395, opened 2021-07-06
ERM-1001: added permission specific to downloading files

EthanFreestone:


aditya-matukumalli (changes requested):
@EthanFreestone Can we not just use the `IfPermission` component here ([sample usage](https://github.com/folio-org/stripes-erm-components/blob/master/lib/ViewOrganizationCard/ViewOrganizationCard.js#L114) instead of wrapping the nested app level components `withStripes` and hence avoid all the prop drilling at the app level? Am I missing something?

EthanFreestone:
> @EthanFreestone Can we not just use the `IfPermission` component here ([sample usage](https://github.com/folio-org/stripes-erm-components/blob/master/lib/ViewOrganizationCard/ViewOrganizationCard.js#L114)) instead of wrapping the nested app level components `withStripes`. That way we could avoid all the prop drilling at the app level? Am I missing something?

My main concern was that this is an "if A then B,else C" scenario, so using IfPermission rather than a switch on the boolean hasPerm seems a bit overkill. That way we'd have to check ifPermission, then render the link or check if anything rendered and if not render a string.


In addition we're doing this in a shared component, so rather than having the permission string passed, which felt a bit dodgy, I thought it preferable to pass a boolean directly which controls the link vs string.

As it stands though, the app drilling is in stripes-erm-components, meaning that the only thing the apps need concern themselves with is how to set up that boolean, in this case by grabbing the stripes object and performing a "hasPerm".

I'm not really sure we can avoid prop drilling in stripes-erm-components, either its a boolean control or its the name of the permission we're checking for right?. Boolean control felt more extensible and useful.

EthanFreestone:
If it's preferable to a bunch of `withStripes` wrapped components, I can rewrite them all to functional and use the useStripes hook?

aditya-matukumalli (approved):

PR #423, opened 2021-09-30
ERM-1700 Apply a four column layout to the User card for agreements a…

MonireRasouli:
Link the user "Name" field to the user record in the Users app.
Apply a four column layout to the the card.


lib/InternalContactsFieldArray/UserField.js
```diff
@@ -107,14 +108,21 @@ export default class UserField extends React.Component {
       email,
       phone
     } = this.props.user?.personal ?? {};
+    const { user } = this.props;
+    const id = get(user, 'user.id');
```
aditya-matukumalli, line 112: Avoid using get. Use optional chaining instead

lib/InternalContactsFieldArray/UserField.js
```diff
@@ -107,14 +108,21 @@ export default class UserField extends React.Component {
...
     return (
       <div data-test-user-card>
         <Row>
-          <Col md={5} xs={12}>
+          <Col md={3} xs={12}>
             <KeyValue label={<FormattedMessage id="stripes-erm-components.contacts.name" />}>
               <span data-test-user-name>
-                {renderUserName(this.props.user)}
+                <Link
+                  data-test-contact-link
```
aditya-matukumalli, line 121: Do we use this data-test-contact-link attribute somewhere?

lib/InternalContactsFieldArray/UserField.js
```diff
@@ -125,7 +133,7 @@ export default class UserField extends React.Component {
               </span>
             </KeyValue>
           </Col>
-          <Col md={4} xs={6}>
+          <Col md={6} xs={6}>
```
aditya-matukumalli, line 136: I think you can avoid passing both xs and md when they are the same value. Please double check

aditya-matukumalli (changes requested):

aditya-matukumalli (approved):

PR #463, opened 2022-03-07
chore: RTL

EthanFreestone:
Set up RTL environment for jest testing

ERM-2041


tests/helpers/translationsProperties.js
```diff
@@ -0,0 +1,39 @@
+import translations from '../../translations/stripes-erm-components/en';
+
+const translationsProperties = [
+  {
+    prefix: 'ui-erm-comparisons',
```
CalamityC, line 5: shouldn't this be `stripes-erm-components` ?

PR #466, opened 2022-03-21
ERM-2051 primary terms are listed in drop-down for optional terms

CalamityC:
* add filter for non-primary custom properties in renderCustomPropertyName function


EthanFreestone (changes requested):
The change looks good, but I don't want to bring in the testing dep change as a part of this commit

CalamityC:
> The change looks good, but I don't want to bring in the testing dep change as a part of this commit

But otherwise we have a lint warning ...

PR #482, opened 2022-05-11
ERM-1315 deprecate components

MonireRasouli:


CalamityC:
@MonireRasouli I'm a bit confused: You're deprecating `CustomPropertyFieldEdit` and `CustomPropertyFieldView`, but both components are imported by `CustomPropertyField` which you're not deprecating. I find `CustomPropertyField` is being imported only by `CustomPropertiesConfigListFieldArray` which is also deprecated. So the correct action for my understanding would be to deprecate also `CustomPropertyField`. Please double check that @MonireRasouli and @EthanFreestone 
Also if you're writing the console warnings make sure that the corrrect components are correctly named in the warning, it seems that it's always `CustomPropertyField` ("Warning: <CustomPropertyField> is deprecated ...") in all deprecated components.
And shouldn't we mention the deprecated components in CHANGELOG.md, @EthanFreestone ?

EthanFreestone:
> @MonireRasouli I'm a bit confused: You're deprecating `CustomPropertyFieldEdit` and `CustomPropertyFieldView`, but both components are imported by `CustomPropertyField` which you're not deprecating. I find `CustomPropertyField` is being imported only by `CustomPropertiesConfigListFieldArray` which is also deprecated. So the correct action for my understanding would be to deprecate also `CustomPropertyField`. Please double check that @MonireRasouli and @EthanFreestone Also if you're writing the console warnings make sure that the components are correctly named in the warning, it seems that it's always `CustomPropertyField` ("Warning: <CustomPropertyField> is deprecated ...") in all deprecated components. And shouldn't we mention the deprecated components in CHANGELOG.md, @EthanFreestone ?

Agreed to all the above, sending back to you @MonireRasouli, thank you @CalamityC

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change history for stripes-erm-components
 
+## 2022-05-11
+* ERM-1315 Deprecate `<CustomPropertiesConfigListFieldArray>`, `<CustomPropertyField>` , `<CustomPropertyFieldEdit>` and `<CustomPropertyFieldView>`
+
 ## 6.2.0 In progress
 * ERM-2095 Replace babel-eslint with @babel/eslint-parser
 * Deprecate `<Card>`, `<CustomPropertiesList>` in preparation for removal in a future release
```
CalamityC, line 8: We usually place our new entries in the `In progress` section and don't make entries per date. I would place line 4 underneath line 8. Sorry for being nitpicking

EthanFreestone: @MonireRasouli, sorry to push this back again, but I agree with Claudia above :)

EthanFreestone (changes requested):

CHANGELOG.md
MonireRasouli: no problem, it fixed

EthanFreestone (approved):

PR #520, opened 2022-07-18
build: Added dev dep

EthanFreestone:
Added dev dep for babel preset-env, which we use in our babel config


EthanFreestone:
Thought I needed to make changes to get a ui-licenses PR through, possible that we don't so converting to draft to test

PR #540, opened 2022-08-26
ERM-2315 Display all identifiers in a namespace for a resource

CalamityC:


lib/getResourceIdentifier/getResourceIdentifier.js
```diff
@@ -1,7 +1,13 @@
-export default (eresource = {}, type) => {
+export default (eresource = {}, type, returnAll) => {
   const identifiers = eresource?.identifiers || [];
-  // find first identifiers value of namespace 'type'
-  const value = identifiers.find(i => i.identifier?.ns.value === type)?.identifier?.value;
 
-  return value;
+  if (returnAll === true) {
+    // find all identifiers value of namespace 'type', return an array
+    const value = identifiers.filter(i => i.identifier?.ns.value === type).map(r => r?.identifier?.value);
```
EthanFreestone, line 6: You can return this value immediately instead of assigning it to a variable. SonarLint should be catching this kind of thing if you have your IDE configured to show you those

lib/getResourceIdentifier/getResourceIdentifier.js
```diff
@@ -1,7 +1,13 @@
...
+    return value;
+  } else { // returnAll is 'false' or undefined
+    // find first identifiers value of namespace 'type', return a string
+    const value = identifiers.find(i => i.identifier?.ns.value === type)?.identifier?.value;
```
EthanFreestone, line 10: Ditto above

lib/getResourceIdentifier/testResources.js
```diff
@@ -0,0 +1,120 @@
+const eresourceOne = {
```
EthanFreestone, line 1: I'm aware this is the pattern we have at the minute in ERM, but I'm really not a fan of it.


I would prefer that we have a set of exports from the `test/jest` directory which we can import so that we have a consistent set of data to draw from in our tests, rather than setting up `eresource` in 5 places. If our API changes then we have a LOT of places to change this at the moment, instead of one, and it will lead to us missing regressions in the tests.


Please can you set up a file named something reflecting which endpoint it's "mocking" in `test/jest`, and export that/import it into your test as a list, like we'd get back from the endoint?

EthanFreestone (approved):

lib/getResourceIdentifier/getResourceIdentifier.js
CalamityC: >  if you have your IDE configured to show you those

Can you tell me please how should the configuration look like?

lib/getResourceIdentifier/testResources.js
CalamityC: I'm afraid I don't undestand what you mean.

EthanFreestone (approved):

PR #563, opened 2022-10-26
ERM-2395: Remove @folio/stripes-testing as direct dependency of stripes-erm-components

EthanFreestone:


BogdanDenis:
Hey @EthanFreestone, removing `@folio/stripes-testing` seems to be causing build error in Rancher builds:
image
https://jenkins-aws.indexdata.com/job/Rancher/job/UI-Build/541/console

In `index.js` we're exporting interactors, which will in turn require `@interactors/html`.
That dependency comes from two places:
- `devDependencies -> @interactors/html`
- `dependencies -> @folio/stripes-testing -> @interactors/html`

If I understand correctly - we are only installing `dependencies` when building environments, so `@folio/stripes-testing` will not  be installed.
So `@interactors/html` will not be installed at all, which causes the build error.

I don't understand, however, why Jenkins reference builds are not failing... perhaps it's installed from some other module
@zburke I seem to remember that we had the same issue with some other module. Can we install `@folio/stripes-testing` at platform level?

zburke:
@BogdanDenis I don't have good context here. Which "Jenkins reference builds" are you referring to? IIUC, when building any module in CI, we simply run `yarn install` which will install that module's deps and dev-deps. When building [`platform-complete#snapshot`](https://github.com/folio-org/platform-complete/tree/snapshot/) that means we only get the dev-deps of platform complete (`@folio/stripes-cli` and some lint things). 

Since there are no tests defined at the platform level, `@folio/stripes-testing` is not a dev-dep there, a configuration I think is correct. IOW, it sounds like whatever repo is at the source of this Rancher build is mis-configured, because it contains tests but does not properly depend on `@folio/stripes-testing`. 

Possibly additionally, (total shot in the dark) `@folio/stripes-testing` is misconfigured and needs to include `@interactors/with-cypress` as a direct-dep, not just a dev-dep, so that when `@folio/stripes-testing` is pulled in as a dep to some repo that includes Cypress tests, `@interactors/with-cypress` is also installed.

BogdanDenis:
@zburke Perhaps you're right, that this issue should be fixed in this module, since it exports interactors and doesn't depend on `@interactors/html`. I just thought that adding it to `platform-complete`'s dependencies would act as an additional safeguard.
But it would also hide the problem at it's root.

@EthanFreestone would it be ok to revert this PR?

EthanFreestone:
I can revert and then re-release, but does that not still leave us with dependencies out of whack, since stripes-testing would remain a full fat dependency in this module?

What's needed for Nolana first and foremost, and then what can I do to improve this situation moving forward?

One idea I've been toying with is to have a devDep on a new module, stripes-erm-testing, which we could then export these kinds of interactors/mocks from. Won't get into Nolana ofc, but could help for Orchid and beyond.

I'm loath to just revert this before we know what needs to happen in Nolana, because that would leave a release out of sync

EthanFreestone:
Such a module wouldn't necessarily even need to be tied into FOLIO releases mind, similar to what we do with `@k-int/stripes-kint-components`

zburke:
@EthanFreestone, we also need to remove `@testing-library/react` from the dependencies, which I discovered yesterday when trying to bump this module's version in `platform-complete#snapshot`. I think that'll snuff out these build warnings for good.

zburke:
@EthanFreestone, oooof, I'm feeling slow and just now realizing the full scope of the issue here. I apologize for not reading your comments more carefully earlier. I appreciate you trying to do The Right Thing by formally exporting things designed to be shared into other repos and apologize, again, if this this feels like a "no good deed goes unpunished" situation. 

This situation is double-edged: We have a long, dirty history of allowing ui-apps to reach directly into the private paths of stripes-core and stripes-components to gain access to shared testing infrastructure. Eww. OTOH, test infrastructure is just that, and it definitely shouldn't be included in a package's exports. It'll get pruned if tree-shaking is properly configured when building the production bundle (which appears to be the case here 😌) but that doesn't feel very clean either. My inclination is the same as yours: either provide an alternative repo like `@folio/stripes-erm-testing` or move these exports to `@folio/stripes-testing`. In my fantasy world, we convert all our stripes-* repos to mono-repos that export two packages, one of production components and one of test components, but we're not there yet. 

In the short term, my inclination is to remove the test-infrastructure exports here and move them into stripes-testing. Alternatively/temporarily we could let other packages' tests import them directly via private paths, though I'd really rather not perpetuate that gross practice. What do you think?

EthanFreestone:
It's a tricky one. We have on our docket the cleaning up of our whole testing situation. I hate the current imports we're making in each test, and would like to make proper use of jest's manual mocking features (We have examples for this already in `ui-oa`).

However that doesn't really solve the issue of our main point of contact for testing being a library we need a direct dep on.
My inclination is that `stripes-erm-testing` would be the way to go here, and to be honest I'm not sure moving all of our exports and test imports over to `stripes-testing` is any more short term than to `stripes-erm-testing`, they're both pretty large scale changes.

Does this need to be fixed in a Nolana release? If so then this would require a re-release of all of our modules, either pulling in the new exports from stripes-testing or from stripes-erm-testing...

This is already one of our top priorities for Orchid. Has something changed significantly that we can't revert this PR for Nolana and then work on fixing all these imports for then?

PR #595, opened 2023-02-13
ERM-2632, Add a copy icon to copy the reference text used in the docu…

MonireRasouli:
…ment type/reference MCL


EthanFreestone (changes requested):
Correct me if I'm wrong @ostephens but I think we only wanted the copy icon here for the URL, not for all elements in row?

EthanFreestone (approved):

PR #597, opened 2023-02-16
 ERM-2634, If an agreement or license has >10 contacts they do not al…

MonireRasouli:


MonireRasouli:
Behaviour needs to be refactored. The query variable is either a string or an empty array, we need to refactor that at some point to always be a string.

EthanFreestone (approved):

PR #637, opened 2023-09-20
ERM-3040 Dashboard: "New" badge is wrapping onto two lines

CalamityC:
* add overflowX: 'hidden' style to InfoBox in NewBox


CalamityC:
Do not solve this with css but with column width

PR #680, opened 2024-04-30
ERM-3220 Update pagination mechanisms for MCLs to work without stats

CalamityC:
* create new hook for fetch with no stats
* add hasNextPage prop to usePrevNextPagination


lib/hooks/useFetchCurrentAndNext.js
```diff
@@ -0,0 +1,57 @@
...
+import { useQueries } from 'react-query';
+import { useOkapiKy } from '@folio/stripes/core';
+import { generateKiwtQueryParams } from '@k-int/stripes-kint-components';
+
+const defaultParams = { page: 1 };
+
+const useFetchCurrentAndNext = ({
+  params = {},
+  path = '',
+  keyArray = [],
```
EthanFreestone, line 11: I'd like to see this changed out for the ability to _fully_ customise the queryKey. As things stand you're locked to path/paramsObj/whatever keys you add.

Something like `getQueryKey` which defaults to
```
getQueryKey({paramsObj, pathStr, pageNum}) => [pathStr, pageNum, paramsObj]
```
where the current page and next page obviously pass in the correct props to the getQueryKey function.


This allows the calling code to be really specific about setup, in ERM's case we have a (loose) pattern of
`['ERM', <RelevantObject>, <RelevantId>, ..., path, params]`, which would allow us to force a refetch of all ERM objects should we wish, but _also_ we could, say, force a refresh of `['ERM', 'Agreement', <agreementId>]` when we update an agreement, causing _also_ fetches for ` ['ERM', 'Agreement', <agreementId>, 'AgreementLine', <agreementLineId>]` to refresh so they have up to date information...


Basically it should be up to the calling code how the queryKey is set up, since different implementations will have different requirements for the queryKey heirachy, and I don't believe the current implementation has enough flexibility (I appreciate that this may also be a problem in other hooks...)

EthanFreestone (changes requested):
Overall code looks really good. Happy with the overall idea, just one small tweak requested to the `queryKey` logic in useFetchCurrentAndNext.

I think it's probably out of scope for this piece of work, but it'd be kinda neat to not limit `useCurrentAndNext` to just fetching 2 pages. Can it be renamed to something like "useFetchMultiplePages"? For now the logic can remain the same, but with a comment explaining that it's named like this to allow for future expansion to n pages down the line

EthanFreestone (approved):

PR #709, opened 2024-11-07
ERM-3395, Complete absent test cases for stripes-erm-components document filter

MonireRasouli:


lib/DocumentFilter/DocumentFilter.test.js
```diff
@@ -139,7 +126,4 @@ describe('DocumentFilter', () => {
       });
     });
   });
-
```
EthanFreestone, line 142: This todo hasn't been done yet, so shouldn't be removed

lib/DocumentFilter/DocumentFilterField.test.js
```diff
@@ -1,96 +1,98 @@
...
-    });
+      it('displays value label(s)', () => {
+        const { queryAllByText } = renderComponent;
+        expect(queryAllByText('Value')).toHaveLength(expectedFields);
+      });
 
-    if (expectedFields > 0) {
-      test('renders the add rule button', async () => {
-        await Button('Add rule').exists();
+      it('renders DocumentFilterRule component with the correct number of times by clicking the "Add rule" button', async () => {
```
EthanFreestone, line 78: This is a little overcomplicated. The big if/else isn't needed here. We can instead test:

- Does the right number of DocumentFilterRule exist? (Use queryAllByText so the 0 case works)
- Does add rule exist
- Describe clicking add rule
  - Has number of DocumentFilterRules gone up by one?

EthanFreestone: In general the point of the `.each` at the top is so that this kind of case by case logic _isn't_ needed inside the tests themselves. Let me know if you wan tme to walk through this case with you cos it's an interesting shift in test writing

lib/DocumentFilter/DocumentFilterFieldArray.test.js
```diff
@@ -1,53 +1,87 @@
...
       </MemoryRouter>,
       translationsProperties
     );
   });
 
-  test('renders the document filters button', async () => {
+  test('adding/removing filterd fields using the add/remove works as expected', async () => {
+    const { getByText, queryAllByTestId } = renderComponent;
     await Button('Add filter').exists();
+    await waitFor(async () => {
```
EthanFreestone, line 32: Please don't do this in a test.

Describe is for actions, test is for outcomes

lib/DocumentFilter/DocumentFilterFieldArray.test.js
```diff
@@ -1,53 +1,87 @@
...
+
+    await waitFor(async () => {
+      expect(getByText('DocumentFilterField')).toBeInTheDocument();
+    });
+
+    await waitFor(async () => {
+      expect(queryAllByTestId(/document-filter-field\[.*\]/i).length).toEqual(1);
+    });
+
+    await waitFor(async () => {
```
EthanFreestone, line 44: Same here, this is a sequence of describes/tests, not one big lump.

Jest doesn't handle asynchronicity that well across differing environments, so try to keep a test to a single thing, and use describes etc to set up the state that we're testing.

Again let me know if you want an insight into what I mean on a call etc

lib/DocumentFilter/DocumentFilterFieldArray.test.js
```diff
@@ -1,53 +1,87 @@
...
-  // Test card delete removes an item
-  // Test card heading shows right label
-  // Test multiples are separated by OR (check this scales with right number of ORs for 2/3/4 items)
+  test('multiple filters are separated by "OR"card heading shows the right labels', async () => {
+    const { getByText, getByRole, queryAllByTestId } = renderComponent;
+    await Button('Add filter').exists();
+
+    // adding multiple filters
+    await waitFor(async () => {
+      await Button('Add filter').click();
```
EthanFreestone, line 67: Same as above

lib/DocumentFilter/DocumentFilterForm.test.js
```diff
@@ -148,6 +131,12 @@ describe('DocumentFilterForm', () => {
         });
       });
 
+      it('should open the document filter builder modal when the edit button is clicked', () => {
+        const { getByText } = renderComponent;
+        fireEvent.click(getByText('Edit document filters'));
```
EthanFreestone, line 136: Why is this using fireEvent instead of interactors? 

I like the problem solving you've done behind the scenes to make sure this works, but at the bare minimum there needs to be a comment explaining why this case is different. My gut says that a difference in behaviour means either our component is wrong or the interactors are wrong... it's rare that we actually _should_ be using things outside of them.


Again, I like the implication that a bunch of work and problem solving has gone into this though, so kudos! Just worried that a "special case" when not labelled with a comment ends up becoming copied and pasted elsewhere blindly and we end up with 10 ways of doing the same thing.

lib/DocumentFilter/DocumentFilterForm.test.js
```diff
@@ -169,6 +158,25 @@ describe('DocumentFilterForm', () => {
...
+          });
+        });
+
+        it('should close the filter form when cancel button is clicked', async () => {
+          const { queryByText } = renderComponent;
+          await waitFor(async () => {
+            await Button('Edit document filters').click();
+          });
+          await waitFor(async () => {
+            await Button('ui-test-implementor.cancel').click();
```
EthanFreestone, line 173: Describes not tests again pls

lib/DocumentFilter/DocumentFilterRule.test.js
```diff
@@ -51,52 +54,65 @@ describe('DocumentFilterRule', () => {
     await Button('Submit').exists();
   });
 
-  it('renders Name option', () => {
+  test('renders the correct options for the Attribute dropdown', async () => {
```
EthanFreestone, line 57: I separated these deliberately... I want each test to be a single granular check. If this fails we don't know which field broke.

lib/DocumentFilter/DocumentFilterRule.test.js
```diff
@@ -51,52 +54,65 @@ describe('DocumentFilterRule', () => {
...
-    TextArea('Category').exists();
-  });
-
-  it('renders is option', () => {
+  test('renders the correct options for the Comparator dropdown', async () => {
     TextArea('is').exists();
-  });
-
-  it('renders containes option', () => {
     TextArea('containes').exists();
```
EthanFreestone, line 69: This is a typo, not sure where that typo manifests

lib/DocumentFilter/DocumentFilterRule.test.js
```diff
@@ -51,52 +54,65 @@ describe('DocumentFilterRule', () => {
...
-  it('renders File name option', () => {
     TextArea('File name').exists();
   });
 
-  it('renders Category option', () => {
-    TextArea('Category').exists();
-  });
-
-  it('renders is option', () => {
+  test('renders the correct options for the Comparator dropdown', async () => {
```
EthanFreestone, line 67: Same thing as above though... pleasee keep individual checks in their own test/it

lib/DocumentFilter/DocumentFilterRule.test.js
```diff
@@ -51,52 +54,65 @@ describe('DocumentFilterRule', () => {
...
+      );
+    });
+
+    test('renders the submit button', async () => {
+      await Button('Submit').exists();
+    });
+
+    test('removing rules using the remove button works as expected', async () => {
+      const { getByRole } = renderComponent;
+      expect(getByRole('button', { name: /remove rule 2/i })).toBeInTheDocument();
```
EthanFreestone, line 106: Why is this not using Button interactor? Same as above, if there's a reason there needs to be a comment

lib/DocumentFilter/DocumentFilterRule.test.js
```diff
@@ -51,52 +54,65 @@ describe('DocumentFilterRule', () => {
...
+    });
+
+    test('if onDelete function will be called', async () => {
+      const deleteButton = screen.getByLabelText('Remove rule 2');
+      expect(deleteButton).toBeInTheDocument();
+      await waitFor(async () => {
+        await userEvent.click(deleteButton);
```
EthanFreestone, line 113: Same as above, Button interactor is the standard

EthanFreestone (changes requested):

EthanFreestone (approved):

PR #728, opened 2025-05-13
ERM-3704, Hide tags icon when tags are disabled

MonireRasouli:


EthanFreestone:
This has now been merged with test coverage. One thing to note on this is that it effectively sets up the "Tags" button for anything that renders this hook, but it doesn't allow for us to separately render different helpers for agreements/licenses. 

It would allow us to set up the same helpers for both and then simply not render the button to turn those on/off, but this isn't necessarily the best way to go.


It may be valuable in future to have a "default" button implementation, perhaps with "label", "icon" and "badge", somehow? Would provide us with a way to override wshould we wish in the output from useHelperApp, but not necessarily have to do all the work when we have multiple helper apps. Doesn't seem massively useful right now though



This PR also introduces a test for useErmHelperApp, which mocks the underlying useHelperApp. That is _somewhat_ implying that useHelperApp underneath has been properly mocked, which is I believe not the case currently

PR #737, opened 2025-09-12
ERM-3780 Add Acquisition Units to Agreement Line view screen

CalamityC:
* add sort function to policies table (ref. ERM-3779)


lib/AccessControl/View/PoliciesTable/PoliciesTable.test.js
```diff
@@ -70,33 +90,48 @@ describe('PoliciesTable', () => {
...
 
         test('onRemove got called with the correct data', async () => {
           await waitFor(() => {
-            expect(onRemove).toHaveBeenCalledWith(expectedData);
+            expect(onDoRemove).toHaveBeenCalledWith(getExpected());
           });
         });
       });
+
+      // NEW: Clicking "Remove" header changes the sort column (branch meta.name !== sortOrder)
```
EthanFreestone: I will pick this up since @CalamityC is off Monday/Tuesday, but this is not our pattern for Jest tests.

Describe/beforeEach does action, then test tests outcome.

This is a bit of an unwieldy pattern I know, but it is important for readability and also for consistency when running in various ci pipelines.


I think about it in a similar vein to some other testing frameworks with a "when X, then Y" pattern. Actions are not the test, you're testing the reaction of the component to the action.

In addition this will throw `act` exceptions, and so we wrap in a `waitFor` to avoid that (Also a cumbersome pattern, but it's better to avoid the errors and be able to cleanly see the test output -- IIRC it also creates issues in CI  if we don't)

EthanFreestone: We shouldn't litter the tests with `// NEW` comments, because they won't always be NEW

lib/AccessControl/View/PoliciesTable/PoliciesTable.js
```diff
@@ -14,6 +14,30 @@ const PoliciesTable = ({
 }) => {
   const intl = useIntl();
 
+  // --- 1. Track sorting state ---
+  const [sortOrder, setSortOrder] = useState('name');
+  const [sortDirection, setSortDirection] = useState('ascending');
+  const nonInteractiveHeaders = ['description', 'restrictions'];
```
EthanFreestone: Remove should also be not-sortable (which will change tests)

lib/AccessControl/View/PoliciesTable/PoliciesTable.test.js
EthanFreestone: Hmmm... actually the act errors seem to be gone. I'm going to remove from here for now and hope that those are no longer required... that'd be nice

lib/AccessControl/View/PoliciesTable/PoliciesTable.test.js
```diff
@@ -49,6 +57,7 @@ describe('PoliciesTable', () => {
       await MultiColumnListHeader(colName).exists();
     });
 
+    // still keep your explicit content checks
```
EthanFreestone, line 59: This isn't a useful comment

lib/AccessControl/View/PoliciesTable/PoliciesTable.test.js
```diff
@@ -22,11 +23,18 @@ describe('PoliciesTable', () => {
...
-      testTitle: 'with allowRemove: :false',
+      testTitle: 'with allowRemove: :true',
     },
   ])('$testTitle', ({ allowRemove, onRemove }) => {
+    let sortedByNameAsc;
+    let sortedByNameDesc;
+
     beforeEach(() => {
-      renderComponent = renderWithIntl(
+      // Mirror PoliciesTable's default sort: by policy.name, ascending
```
EthanFreestone, line 32: This isn't a useful comment

PR #771, opened 2026-05-18
ERM-3994, Add TagFilter shared component

MonireRasouli:


lib/TagsFilter/TagsFilter.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
EthanFreestone, line 1: Not going to block because of this, but we no longer need to import React at the top of our components

EthanFreestone (approved):

PR #775, opened 2026-06-30
ERM-4076, Agreements and Licenses UIs should only do access checks if…

MonireRasouli:
* useCanAccess, usePolicies, and useRestrictionPolicies: Default doAccessControl to false and refactor useQuery to use the object signature.
* update the tests
* ADD README files for the hooks


lib/AccessControl/hooks/useCanAccess/useCanAccess.js
```diff
@@ -4,23 +4,41 @@ import { useQuery } from 'react-query';
...
+  accessControlEndpoint = null,
   resourceEndpoint, // Should be the BASE endpoint, so erm/sas for example
   resourceId, // Optional, since CREATE works on base endpoint
   restriction,
   queryNamespaceGenerator,
   queryOptions = {}
 }) => {
   const ky = useOkapiKy();
+
+  const inferredAccessControlEndpoint = useMemo(() => {
```
EthanFreestone: This isn't correct.

The main canAccess will _still_ need to use resource endpoint

lib/AccessControl/hooks/useCanAccess/useCanAccess.js
```diff
@@ -4,23 +4,41 @@ import { useQuery } from 'react-query';
...
+    return accessControlEndpoint ?? (resourceEndpoint ? resourceEndpoint.split('/')[0] + '/accessControl' : null);
+  }, [accessControlEndpoint, resourceEndpoint]);
+
+  const { enabledEngines, isLoading: isEnabledEnginesLoading } = useEnabledEngines({ endpoint: inferredAccessControlEndpoint });
```
EthanFreestone: I'm not sure I want an extra lookup in this hook, that's making some assumptions about behaviour (Albeit quite good ones right now)


We have a hook `useGetAccess` which is our actual "all in one" query hook, and I think that should probably be the place for an additional lookup.


We ALSO have `useDoAccessControl` which already provides a single boolean answer, we can extend that to expose a "isDoAccessControlLoading", and then use the queryOptions to pass

```
{
  enabled: !isDoAccessControlLoading || !doAccessControl
}
```

so that the canAccess lookup doesn't happen for enabled engines false.


We ALSO need to ensure that that sets to "true" so that it's ignored, not failing shut.

lib/AccessControl/hooks/usePolicies/usePolicies.js
```diff
@@ -12,13 +16,23 @@ const usePolicies = ({
 }) => {
   const ky = useOkapiKy();
 
+  const inferredAccessControlEndpoint = useMemo(() => {
```
EthanFreestone: Again, do not infer between them, we have one endpoint for the check, one for the lookup

lib/AccessControl/hooks/usePolicies/usePolicies.js
```diff
@@ -12,13 +16,23 @@ const usePolicies = ({
...
+    return accessControlEndpoint ?? (resourceEndpoint ? resourceEndpoint.split('/')[0] + '/accessControl' : null);
+  }, [accessControlEndpoint, resourceEndpoint]);
+
+  const { enabledEngines, isLoading: isEnabledEnginesLoading } = useEnabledEngines({ endpoint: inferredAccessControlEndpoint });
```
EthanFreestone: Should use doAccessControl, see above

EthanFreestone (changes requested):
I think this is not quite right, please make changes AND add README files to each hook explaining when to use each one, and what they do, what assumptions are being made etc

EthanFreestone (changes requested):
Sorry @MonireRasouli but this is not quite right.


Firstly, we shouldn't be passing "do access control" and loading states into each hook... the hook accepts an enabled via the queryParameters and it is up to the _caller_ to handle that.

useCanAccess is NOT used anywhere directly in our modules, so it won't matter, but if it were then it would be up to the MODULE to run a doAccessControl and pass that in via enabled.


useGetAccess have an option to use doAccessControl, and controlling the enabled prop directly via that.

However it is NOT an all in one and should _not_ be returning policies I don't think.


usePolicies and useRestrictionPolicies should have options to handle the doAccessControl call first internally



In the future it might be preferable to do doAccessControl via a context or something instead, as this call won't change very often, so leaving it optional (but defaulting to on) is the way.



I will pick this up for now and get the changes in

